PR #1, opened 2020-01-07
Show a breadcrumb when rooted at a lower node

tstirrat:
Screen Shot 2019-12-20 at 12 12 36 PM

### DIFF
--- a/src/components/DataViewer/NodeParentKey.tsx
+++ b/src/components/DataViewer/NodeParentKey.tsx
@@ -1,0 +1,88 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


src/components/DataViewer/NodeParentKey.tsx
```diff
@@ -0,0 +1,88 @@
...
+  navigateToRef: (ref: firebase.database.Reference) => void
+) {
+  let ref: firebase.database.Reference | null = activeRef;
+  const refs = [];
+  while (ref !== null) {
+    refs.push(ref);
+    ref = ref.parent;
+  }
+
+  const crumbs = refs.reverse().reduce<JSX.Element[]>((crumbs, ref, i) => {
```
yuchenshi: nit: This is imperatively manipulating the `crumbs` array, and I think we should just make it clear by not using reduce which implies immutable reduction. I'd suggest:

```js
const crumbs = [];
for (let i = refs.length - 1; i >= 0; i--) {
  // ... crumbs.push etc
}
```

It's more straightforward to understand and doesn't secretly reverse `refs` in place.

tstirrat: Done

src/components/DataViewer/NodeParentKey.tsx
```diff
@@ -60,7 +60,9 @@ function renderBreadcrumbs(
   navigateToRef: (ref: firebase.database.Reference) => void
 ) {
   const ancestors = getAncestorRefs(activeRef);
-  const crumbs = ancestors.reduce<JSX.Element[]>((crumbs, ref, i) => {
+
+  let crumbs: JSX.Element[] = [];
+  ancestors.forEach((ref, i) => {
```
yuchenshi: nit: This can probably use `for-of` loops. If you do feel strongly about `forEach`, please remove the second argument (the `[]` in `}, []);` on Line 81).

PR #5, opened 2020-01-24
Pack server code together using webpack.

yuchenshi:
To create a zipball:

```bash
npm run build:server
zip dist.zip -r server.bundle.js build

# Now try to extract it somewhere else and see if it runs:
cd /tmp/
unzip /path/to/ui/dist.zip
firebase emulators:exec --project foo --only database 'node server.bundle.js'
```

Now you can visit localhost:3000 for a bundled version of the Emulator UI.

Note that this is not gonna work later if we ever introduce native dependencies (e.g. `grpc`, or anything that depends on `grpc` like Firestore SDK.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 
 # production
 /build
+/server.bundle.js
 
 # misc
 .DS_Store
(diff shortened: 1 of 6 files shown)


samtstern:
Tested building on Linux and using the bundle on both Linux and macOS

PR #8, opened 2020-02-26
New AppBar component to contain top-level navigation

tjlav5:
Some rejiggering of where we load mwc (global index) and the layout
of App to facilitate scrolling/lazy in the future.

Note: global box-sizing change to border-box for general sanity


src/index.scss
```diff
@@ -1,3 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import-normalize;
+
+@import '../node_modules/material-components-web/dist/material-components-web.css';
+
+* {
```
yuchenshi: WDYT about moving this above the `material-components-web.css` import, just in case some styles in `material-components-web.css` rely on box-sizing?

src/components/AppBar/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const AppBar: React.FC<any> = (props) => {
+  const tabs = [defaultLink, ...links].map(({ to, label }) => (
+    <Tab key={label} className="mdc-tab--min-width" {...{ tag: Link, to }}>
+      {label}
+    </Tab>
+  ));
+
+  const activeTab =
+    links.findIndex((l) => props.history.location.pathname.startsWith(l.to)) +
```
yuchenshi: Please use React router [`<NavLink>`](https://reacttraining.com/react-router/web/api/NavLink) instead of manual string matching. (Blocking: please address before this before merging PR).

src/App.tsx
```diff
@@ -29,10 +29,7 @@ const DatabaseRoute: React.FC<{ match: match<{ namespace: string }> }> = ({
 const App: React.FC = () => {
   return (
     <div className="App">
-      <header className="App-header">
-        <h1>Firebase Emulator Suite</h1>
-        <Navigation />
-      </header>
+      <AppBar />
```
yuchenshi: While we're at it, can we also remove `src/components/Navigation/**`? If I read this correctly we don't need my legacy hacky navigation panel.

src/App.scss
```diff
@@ -1,15 +1,8 @@
...
 
-.App-header {
-  width: 30vw;
-  background-color: #282c34;
-  color: white;
-}
-
 .App-main {
-  flex: 1;
-  overflow-y: auto;
```
yuchenshi: Does the main content still scroll? e.g. What happens if you have tons of keys in RTDB Viewer?

src/components/AppBar/index.scss
```diff
@@ -0,0 +1,63 @@
...
+@import '@material/tab/mixins';
+@import '@material/tab-indicator/mixins';
+
+.AppBar {
+  padding: 0 24px;
+
+  .title-grid {
+    align-items: center;
+    align-self: flex-end;
+    display: grid;
```
yuchenshi: Do we really need grid for this one? I kinda understand that it makes layout a little bit easier, but `grid-area` is [not supported in IE](https://caniuse.com/#feat=mdn-css_properties_grid-area). TBH, I think the same layout can be easily achieved using flexbox or even inline blocks, but correct me if I'm wrong.

src/components/AppBar/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+      1 || 0;
+
+  return (
+    <>
```
yuchenshi: Do we need this `<>`?

src/index.scss
tjlav5: Done.

src/App.tsx
tjlav5: Done.

src/App.scss
tjlav5: Yep, so App-main has a height of the viewport minus the app-bar. Each Route-component can set its own scroll pattern... e.g. setting the route-component to height:100%;overflow:scroll; would give you the same effect as today; otherwise the route-component still scrolls but renders the whole node

src/components/AppBar/index.scss
tjlav5: Fair, but are we supporting IE? Last I checked we only support 'last 2 versions' of evergreen browsers. Also, argument for grids, we'd have to add some hacky padding to the subtitle to get the text to line-up or rely on a pseudo-element for the logo in the main title.

src/components/AppBar/index.tsx
tjlav5: Ah not anymore, removed

src/components/AppBar/index.scss
yuchenshi: I think I have a non-hacky non-grid way of doing this but let's not let this block the PR -- let's merge this as-is and we can figure it out later if we have time.

For supported browsers, right now it's configured like this https://github.com/firebase/firebase-tools-ui/blob/486e7146f162663aa74a20797cad8878fdd8f65b/package.json#L41-L45 which I think includes IE11 (it's not "dead" yet). Whether we want that is a different question that we can talk for sure.

yuchenshi:
Tooltip in RTDB viewer seems broken by this PR. I have no idea why though.

src/App.scss
```diff
@@ -1,15 +1,8 @@
 .App {
-  display: flex;
   height: 100vh;
```
yuchenshi:
```suggestion
```

Actually please just drop this line -- I've tested it out and it actually won't do anything visually, and just arbitrarily limits the size of this element while more content just overflows.

Scrolling works out of the box since the whole document is scrollable so we don't need any special handling here.

src/App.scss
```diff
@@ -1,15 +1,8 @@
...
-.App-header {
-  width: 30vw;
-  background-color: #282c34;
-  color: white;
-}
-
 .App-main {
-  flex: 1;
-  overflow-y: auto;
+  height: 100%;
```
yuchenshi:
```suggestion
```

Similarly this line does nothing. Please remove.

src/components/AppBar/index.scss
yuchenshi: How about this? https://github.com/firebase/firebase-tools-ui/pull/9

src/App.scss
tjlav5: Done

src/App.scss
tjlav5: Done

src/App.scss
```diff
@@ -1,15 +1,3 @@
...
-.App-header {
-  width: 30vw;
-  background-color: #282c34;
-  color: white;
-}
-
 .App-main {
-  flex: 1;
-  overflow-y: auto;
+  padding-top: 176px;
```
yuchenshi: This seems quite magical and probably won't work if font sizes are different. As far as I can tell these material components from `rmwc` are all based on `rem`.

However, I don't really know what a proper solution would be. We can use [`<TopAppBarFixedAdjust />`](https://github.com/jamesmfriedman/rmwc/tree/master/src/top-app-bar#topappbarfixedadjust) to fill in the gap for the app bar itself, but we cannot easily reproduce the height of the title text + icon.

Or we can just make everything static and make the `.App-main` scrollable, instead of making the full document scrollable. Thoughts?

src/components/AppBar/index.scss
```diff
@@ -0,0 +1,63 @@
...
+    grid-gap: 4px;
```
yuchenshi:
```suggestion
    grid-gap: .25rem;
```

Ditto

src/components/AppBar/index.scss
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '@material/tab/mixins';
+@import '@material/tab-indicator/mixins';
+
+.AppBar {
+  padding: 0 24px;
```
yuchenshi:
```suggestion
  padding: 0 1.5rem;
```

To align with `rem`s used in the rmwc components.

src/components/AppBar/index.scss
```diff
@@ -0,0 +1,63 @@
...
+      grid-area: subtitle;
+    }
+
+    svg {
+      grid-area: logo;
+    }
+  }
+
+  .mdc-tab-bar {
+    padding-left: 24px;
```
yuchenshi:
```suggestion
    padding-left: 1.5rem;
```

src/components/AppBar/index.scss
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  .mdc-tab {
+    @include mdc-tab-active-text-label-color('secondary');
+    @include mdc-tab-horizontal-padding(5px);
```
yuchenshi:
```suggestion
    @include mdc-tab-horizontal-padding(.25rem);
```

not exact math but should work fine.

src/components/AppBar/index.scss
```diff
@@ -0,0 +1,63 @@
...
+    @include mdc-tab-text-label-color(white)
+
+    &:nth-of-type(n+1) {
+      margin-left: 16px;
```
yuchenshi:
```suggestion
      margin-left: 1rem;
```

src/components/Database/DatabaseDefaultRoute.test.tsx
```diff
@@ -18,20 +18,30 @@ import * as React from 'react';
...
   const { getByText } = render(
-    <MemoryRouter>
-      <Route exact path="/database/sample/data">
-        SUCCESS
-      </Route>
-      <DatabaseDefaultRoute projectId={'sample'} />
-    </MemoryRouter>
+    <MemoryRouter initialEntries={['/']}>
+      <Provider store={store}>
+        <Route exact path="//sample/data">
```
yuchenshi: I'll get to it later...

src/components/Database/DatabaseDefaultRoute.tsx
```diff
@@ -16,21 +16,42 @@
...
 import { AppState } from '../../store';
+import Database from './index';
 
 export interface PropsFromState {
   projectId?: string;
 }
 
 export type Props = PropsFromState;
 
+const DatabaseRoute: React.FC<{ match: match<{ namespace: string }> }> = ({
```
yuchenshi: nit: inline this.

src/App.scss
tjlav5: Pulled in the heights from the underlying mdc-lib, heights from prominent-app-bar and tabs.

I think this is still better than making App-main scrollable as we would lose a nice animation where a drop-shadow is conditionally rendered on the mdc/top-app-bar

src/components/AppBar/index.scss
tjlav5: My thought is that we actually should not use rem's in the core, we use px across the Console and we should stay consistent with that spec. _if_ we want to revisit this, someone could easily come back and write a script to update these all...

src/components/AppBar/index.scss
tjlav5: see comments below on rem vs px

src/components/AppBar/index.scss
tjlav5: see comments below on rem vs px

src/components/AppBar/index.scss
tjlav5: see comments below on rem vs px

src/components/AppBar/index.scss
tjlav5: see comments below on rem vs px

src/components/Database/DatabaseDefaultRoute.tsx
tjlav5: Inline it on :14?

src/App.scss
yuchenshi: Neat!

src/components/Database/DatabaseDefaultRoute.tsx
```diff
@@ -16,21 +16,42 @@
...
   }
-  // Default to the namespace named after the project id.
-  return <Redirect to={`/database/${projectId}/data`} />;
+
+  return (
+    <Switch>
+      <Route exact path={path}>
+        <Redirect to={`${url}/${projectId}/data`} />;
+      </Route>
+      <Route path={`${path}/:namespace/data`} component={DatabaseRoute}></Route>
```
yuchenshi:
```suggestion
      <Route path={`${path}/:namespace/data`} component={({params}) => <Database namespace={params.namespace} />}></Route>
```

src/components/Database/DatabaseDefaultRoute.tsx
yuchenshi: I've added a suggested edit below to clarify.

PR #12, opened 2020-03-04
RTDB Database Picker

yuchenshi:


src/components/Database/Database.scss
```diff
@@ -0,0 +1,16 @@
+.Database {
+  display: flex;
+  background-color: #fff;
```
tjlav5: nit: should we be ordering our css a-z?

src/components/Database/Database.tsx
```diff
@@ -56,7 +56,7 @@ export const Database: React.FC<Props> = ({ config, namespace }) => {
   );
 
   return (
-    <div className="Database">
+    <div className="Database-Database">
       <ThemeProvider options={theme}>
```
tjlav5: Do we need a themeProvider here if there is one higher-up? Is anything changing from the global theme?

src/components/Database/DatabasePicker.scss
```diff
@@ -0,0 +1,16 @@
+.Database-Picker {
```
tjlav5: nit: missing license?

src/components/Database/DatabasePicker.tsx
```diff
@@ -0,0 +1,138 @@
...
+  databases?: string[];
+}
+
+export type Props = PropsFromState & {
+  primary: string;
+  current: string;
+  navigation: (db: string) => ReactElement;
+};
+
+const databaseIcon = (
```
tjlav5: Do we want this anywhere else? Consider pulling out into /icons.ts?

tstirrat: Agreed, static asset, make it's own file

src/components/Database/DatabasePicker.tsx
```diff
@@ -0,0 +1,138 @@
...
+}> = ({ name, navigation, activated }) => (
+  <ListItem
+    activated={activated}
+    tag={props => {
+      const element = navigation(name);
+      return { ...element, props: { ...element.props, ...props } };
+    }}
+  >
+    <ListItemGraphic
+      style={{ width: '36px', height: '36px' }}
```
tstirrat: inline object literal consts break purity (because they are created new every render), extract out to module level const

src/components/Database/DatabasePicker.tsx
```diff
@@ -0,0 +1,138 @@
...
+      icon={{
```
tstirrat: same here re literals

src/components/Database/Database.tsx
yuchenshi: Yes, primary color (like modal button) is different. Whether we want that or not is a different question though, but let's keep it like that until we do the polishing rounds later.

src/components/Database/Database.scss
yuchenshi: Sure, I've re-ordered them. In general, I don't think we need to be strict about this, especially because we don't actually have that many properties per rule.

src/components/Database/DatabasePicker.scss
yuchenshi: Ugh somehow google/addlicense does not handle scss files so i added manually. Let me find another tool and also add a check later.

PR #14, opened 2020-03-04
Firestore panels

tjlav5:
* A panel is either a Document/Collection
* If the Document component has no ref, it is treated as the root
* Route matching is exclusively handled by Document/Collection;
  CollectionList is now just for linking

### DIFF
--- a/src/components/Firestore/Document.tsx
+++ b/src/components/Firestore/Document.tsx
@@ -26,41 +26,59 @@ import DocumentPreview from './DocumentPreview';
 import Collection from './Collection';
 import { useApi } from './ApiContext';
 
-export interface Props {
-  reference?: firestore.DocumentReference;
-  snapshot?: firestore.DocumentSnapshot;
-}
-
-export const Document: React.FC<Props> = ({ reference, snapshot }) => {
+const Doc: React.FC<{
+  id: string;
+  selectionGetter: (id: string) => firestore.CollectionReference;
+  children: React.ReactNode;
+}> = ({ id, selectionGetter, children }) => {
   const { url } = useRouteMatch()!;
-  const api = useApi();
 
   return (
     <>
       <div className="Firestore-Document">
-        <div className="Firstore-Document-Preview">
+        <div className="Firstore-Document-Header">
           <Typography use="body1">
-            <Icon icon="insert_drive_file" /> {reference && reference.id}
+            <Icon icon="insert_drive_file" /> {id}
           </Typography>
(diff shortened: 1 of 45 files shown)


src/components/Firestore/Document.tsx
```diff
@@ -19,28 +19,47 @@ import './index.scss';
...
+import { Route, useRouteMatch } from 'react-router-dom';
 
 import CollectionList from './CollectionList';
 import DocumentPreview from './DocumentPreview';
+import Collection from './Collection';
+import { useApi } from './ApiContext';
 
 export interface Props {
-  reference: firestore.DocumentReference;
+  reference?: firestore.DocumentReference;
```
yuchenshi: This feels hacky to me and introduces a whole bunch of conditionals in the render body. For example, the type system doesn't reflect the fact that we'll never have `snapshot` but not `reference`, and if we ever have a null error it can be hard to debug since it now shows as root.

Let's actually make two different components, one for `<DatabaseRoot>` and another for `<Document>` and extract the common things include routes into a lower-level component. This will also greatly simply future work with missing documents.

tjlav5: I could type this to be more narrow... no-reference/no-snapshot, reference,maybe-snapshot

Or do you think HOCs are the way to go?

yuchenshi: I think we can get this done by extracting helper components that are present in both root and document. That component can probably leave `children` open for data viewer etc. I think that will be cleaner and less mental overhead than messing with typing. I don't see an immediate need for HOC, but feel free to prove me wrong.

tjlav5: Ack... I updated Document.tsx, take a look. There is now Root/Document components that use an underlying private Doc.

src/components/Firestore/Document.tsx
```diff
@@ -26,41 +26,59 @@ import DocumentPreview from './DocumentPreview';
...
 
-export interface Props {
-  reference?: firestore.DocumentReference;
-  snapshot?: firestore.DocumentSnapshot;
-}
-
-export const Document: React.FC<Props> = ({ reference, snapshot }) => {
+const Doc: React.FC<{
+  id: string;
+  selectionGetter: (id: string) => firestore.CollectionReference;
```
yuchenshi: nit: rename to `collectionById`? The current name is confusing to me.

tjlav5: Done

PR #17, opened 2020-03-05
Update to latest purple theme.

yuchenshi:
image


samtstern:
Damn this looks :fire: :fire: :fire:

src/index.tsx
```diff
@@ -53,8 +53,7 @@ ReactDOM.render(
     <ThemeProvider
       options={{
         background: '#F6F7F9',
-        primary: '#0C2D48',
-        secondary: '#008B8B',
+        primary: '#681DA8',
```
tjlav5: Let's move all of these color vars out into a color.ts file...

yuchenshi: Done.

PR #18, opened 2020-03-05
Revamp Overview again according to new design.

yuchenshi:
image

### DIFF
--- a/src/components/Home/index.tsx
+++ b/src/components/Home/index.tsx
@@ -23,6 +23,7 @@ import { Typography } from '@rmwc/typography';
 import { Grid, GridCell } from '@rmwc/grid';
 import { Card, CardActionButton, CardActionIcons } from '@rmwc/card';
 import { ListDivider } from '@rmwc/list';
+import { Link } from 'react-router-dom';
 
 export interface PropsFromState {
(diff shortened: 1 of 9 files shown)


samtstern:
I don't know where the string comes from but it says "local enviornment" not "local environment"

We could leave that in as PR bait though :-)

src/components/Home/index.tsx
```diff
@@ -19,13 +19,54 @@ import { connect } from 'react-redux';
...
+        <Typography use="headline6" tag="h3">
+          {name}
+        </Typography>
+        <Typography use="subtitle2" tag="div">
+          Status
+        </Typography>
+        <Typography use="headline6" tag="div">
+          On
+        </Typography>
+        <br />
```
tjlav5: nit: don't use breaks for styling

yuchenshi:
> I don't know where the string comes from but it says "local enviornment" not "local environment"
> 
> We could leave that in as PR bait though :-)

Not introduced by this PR, but I've fixed it anyway.

src/components/Home/index.tsx
yuchenshi: Sure, no breaks now.

PR #19, opened 2020-03-05
Firestore panel headers

tjlav5:
image

### DIFF
--- a/src/components/Firestore/FirestoreLogo.tsx
+++ b/src/components/Firestore/FirestoreLogo.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019 Google LLC
+ * Copyright 2020 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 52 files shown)


src/index.tsx
```diff
@@ -51,7 +51,7 @@ ReactDOM.render(
     // Globally disable ripples
     ripple={false}
   >
-    <ThemeProvider options={{ background, primary }}>
+    <ThemeProvider options={{ background, primary, secondary: '#F5F5F5' }}>
```
yuchenshi: Move this also into `colors`?

yuchenshi: Also, thoughts on saving secondary for the `#476282` color? It's use in quite a few places like overview text and banners in RTDB.

src/components/Firestore/FirestoreLogo.tsx
```diff
@@ -0,0 +1,44 @@
+/**
+ * Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
tjlav5: ha! copy pasta, thx

src/index.tsx
tjlav5: Ack, good call... moved to colors and also isolated this surface within panel header

PR #21, opened 2020-03-06
Firestore clear all

tjlav5:
Add the top-level clear-all functionality for Firestore

image


src/components/Firestore/api.tsx
```diff
@@ -36,16 +37,22 @@ export default class DatabaseApi implements FirestoreApi {
...
     this.baseUrl = `http://${config.hostAndPort}/v1/projects/${projectId}/databases/${databaseId}/`;
+    this.baseEmulatorUrl = `http://${config.hostAndPort}/emulator/v1/projects/${projectId}/databases/${databaseId}/`;
   }
 
   delete(): Promise<void> {
     return this.cleanup();
   }
 
-  private async restPost(path: string, jsonBody: {}) {
+  private async restPost(
```
yuchenshi: nit: Rename to `restRequest` since it's not just post and also make `method` required to be explicit.

src/components/Firestore/index.test.tsx
```diff
@@ -49,3 +51,49 @@ it('shows the top-level collections', async () => {
...
+  );
+
+  act(() => getByText('clear all data').click());
+
+  await wait();
+
+  expect(nukeSpy).toHaveBeenCalled();
+});
+
+it('does not trigger clearing all data', async () => {
```
yuchenshi:
```suggestion
it('does not trigger clearing all data if dialog is not confirmed', async () => {
```

src/components/Firestore/api.tsx
tjlav5: Done

PR #24, opened 2020-03-11
Use original case on buttons/tabs

tstirrat:
Screen Shot 2020-03-11 at 12 09 01 PM


src/_material-overrides.scss
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// for more specificity over originals
+.mdc-button.mdc-button {
```
yuchenshi: OMG this is ugly and I think we should do this instead:

```suggestion
.App .mdc-button {
```

And move to `App.scss` or `index.scss` if you don't mind.

tstirrat: I think we should keep material overrides in a single place. So there is clear obvious intent where we override the default material things. We may also need to do this for *all* select boxes etc. so on and so forth

tstirrat: Also "OMG this is ugly" could have been omitted ;)

yuchenshi: point taken

tjlav5: Is there any chance we can upgrade our material lib to use the latest mixins which includes text-transform: https://github.com/material-components/material-components-web/blame/master/packages/mdc-tab/_mixins.scss#L212

yuchenshi: It's a transitive dependency from rmwc so I'd say let's try upgrading to latest rmwc and then `npm update`. If it still doesn't get us there (probably won't for 4.x), then I'd not spend too much effort there.

tstirrat: Sadly, its v4+ and RMWC is still on 3.

tstirrat: https://github.com/jamesmfriedman/rmwc/blob/master/src/tabs/package.json#L35

PR #26, opened 2020-03-11
Fix card default radius and elevation

tstirrat:


src/App_material-overrides.scss
```diff
@@ -14,12 +14,18 @@
...
+    text-transform: none;
+  }
 
-// TODO: Use text-transform mixins once RMWC updates to @material/tabs v4.0+
-// https://github.com/material-components/material-components-web/commit/22d7ad2fb1796fbd71a3a5ee9c7ab2e77c60e34e
-.App .mdc-tab__text-label {
-  text-transform: none;
+  // TODO: Use text-transform mixins once RMWC updates to @material/tabs v4.0+
+  // https://github.com/material-components/material-components-web/commit/22d7ad2fb1796fbd71a3a5ee9c7ab2e77c60e34e
+  .App .mdc-tab__text-label {
```
yuchenshi:
```suggestion
  .mdc-tab__text-label {
```

src/App_material-overrides.scss
```diff
@@ -14,12 +14,18 @@
  * limitations under the License.
  */
 
-.App .mdc-button {
-  text-transform: none;
-}
+.App {
```
tjlav5: I wouldn't be opposed to dropping these .App prefixes altogether

yuchenshi: Then you'd have to do something like `.mdc-button.mdc-button` to increase specificity, as discussed in the previous PR. I'd rather have the .App prefix.

tstirrat: We could do `body .x` or we could ensure this file is always loaded later (slightly more risky).. 

... will leave at `.App` for now

yuchenshi: Late to the party, but please don't use `body .x` here, otherwise any more specific rules we write in the future will also need to include at least a tag, leading to weird rules like `body .my-component .mdc-button` or `.my-component div.mdc-button`. I'm happy to talk but I'd still strongly suggest that we keep `.App`.

PR #30, opened 2020-03-12
Dialogs for document/collection creation

tstirrat:
These are very rough (the `fullWidth` styles seem to overlap in RMWC.. odd). I'm electing to get the basics in, then follow up with visuals, validation etc.

They work perfectly™️, except when you try to add invalid JSON data :)

I would much prefer them as a promise interface, compared to the `promptClearAll` there's so much extra state and boilerplate.... but this is good enough for now.

Start a collection (2 step):

Screen Shot 2020-03-12 at 2 50 06 PM
Screen Shot 2020-03-12 at 2 50 13 PM

Add a document (single dialog):

Screen Shot 2020-03-12 at 2 50 23 PM


src/components/Firestore/Collection.tsx
```diff
@@ -34,6 +38,13 @@ export const Collection: React.FC<Props> = ({ collection }) => {
   const [collectionSnapshot, loading, error] = useCollection(collection);
   const { url } = useRouteMatch()!;
 
+  const [isDialogOpen, setDialogOpen] = useState(false);
```
yuchenshi: nit: rename to isAddDocumentDialogOpen

src/components/Firestore/CollectionList.tsx
```diff
@@ -33,13 +37,40 @@ export const CollectionList: React.FC<Props> = ({ reference }) => {
   const api = useApi();
   const collections = useCollections(api, reference);
 
+  const [isDialogOpen, setDialogOpen] = useState(false);
```
yuchenshi: nit: isAddCollectionDialogOpen

src/components/Firestore/Collection.tsx
```diff
@@ -34,6 +38,13 @@ export const Collection: React.FC<Props> = ({ collection }) => {
   const [collectionSnapshot, loading, error] = useCollection(collection);
   const { url } = useRouteMatch()!;
 
+  const [isAddDocumentDialogOpen, setAddDocumentDialogOpen] = useState(false);
```
tjlav5: From what I understand hooks always need to be at the top of the FC, if that's the case I'd actually consider moving this up one line and putting a line break between this hook and addDocument; mostly in the future case that someone accidentally adds something before this hook.

src/components/Firestore/dialogs/AddCollectionDialog.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+interface AddCollectionStepProps {
+  documentRef?: firestore.DocumentReference;
+  onChange: (value: string) => void;
+}
+
+export const AddCollectionStep = memo<AddCollectionStepProps>(
+  ({ documentRef, onChange }) => {
+    const [id, setId] = useState('');
+    const updateId = useCallback(
```
tjlav5: Is this needed? Seems like we can just use a regular function?

src/components/Firestore/dialogs/AddCollectionDialog.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+    const getValue = useCallback(
+      () => ({
+        collectionId,
+        document,
+      }),
+      [collectionId, document]
+    );
+
+    const emitValueAndClose = useCallback(
```
tjlav5: Same commentary, do we need this hook? Are we re-rendering Dialog?

src/components/Firestore/Collection.tsx
tstirrat: Luckily eslint will warn you if you do it wrong, but it's a good suggestion, I'll do it.

PR #31, opened 2020-03-12
Move src/DataViewer -> src/Database/DataViewer

tstirrat:
We aren't reusing any of it in Firestore/Logs


yuchenshi:
WANT_LGTM=@tjlav5

tstirrat:
tj agreed in hangouts

PR #35, opened 2020-03-13
Firestore document preview and basic-editing

tjlav5:
image

Breaking this up into smaller bits... this includes:
- a basic list view
- an editor (can only edit values right now)
- adding field in the preview to maps/arrays
- deleting fields (preview and editor)

Note on structure, reducer function is shared between Field and Editor, but they each have their own Provided reducer instance. In effect: Field will generally exist within the whole document, while Editor will only have its state initialized to the selected Field's value.

Still to come:
- better presentation that JSON.stringify in the ListItem (indentation too)
- changing field type
- specific field type editors
- conditionally changing field-key (especially in document-creation/add-field)


yuchenshi:
Please kindly rebase this.

src/components/Firestore/Field/DocumentStore.tsx
```diff
@@ -0,0 +1,106 @@
...
+import * as actions from './actions';
+
+const reducer = createReducer<FirestoreAny, Action>({})
+  .handleAction(actions.reset, (state, { payload }) => {
+    return payload;
+  })
+  .handleAction(
+    actions.addField,
+    produce((draft: FirestoreAny, { payload }) => {
+      const parent = get(draft, getParentPath(payload.path)) || draft;
```
yuchenshi:
```suggestion
      const parent = get(draft, getParentPath(payload.path), draft);
```

Let's use the defaultValue third param instead of `||` here, which is more predictable with `0`, `[]`, etc.

src/components/Firestore/Field/DocumentStore.tsx
```diff
@@ -0,0 +1,106 @@
...
+      if (isMap(parent)) {
+        parent[getLeafPath(payload.path)] = payload.value;
+      } else if (isArray(parent)) {
+        parent.push(payload.value);
+      }
```
yuchenshi: What happens if parent is string, null, etc.?

src/components/Firestore/Field/DocumentStore.tsx
```diff
@@ -0,0 +1,106 @@
...
+    })
+  )
+  .handleAction(
+    actions.updateField,
+    produce((draft: FirestoreAny, { payload }) => {
+      const parent = get(draft, getParentPath(payload.path)) || draft;
```
yuchenshi: same

src/components/Firestore/Field/DocumentStore.tsx
```diff
@@ -0,0 +1,106 @@
...
+      'useDocumentDispatch must be used within a DocumentProvider'
+    );
+  }
+  return context;
+}
+
+export function useFieldState(path: string[]) {
+  const documentState = useDocumentState();
+  if (path.length === 0) return documentState;
+  return get(documentState, path) || '';
```
yuchenshi: Why empty string though? Why not `undefined`?

src/components/Firestore/Field/Editor.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  useEffect(() => {
+    onChange && onChange(state);
+  }, [onChange, state]);
+
+  return <NestedEditor path={[]} rootKey={rootKey} />;
+};
+
+/**
+ * Field with CTAs for editing as well as rendering applicable child-nodes
```
yuchenshi: What is CTA?

src/components/Firestore/Field/Editor.tsx
```diff
@@ -0,0 +1,108 @@
...
+ */
+const NestedEditor: React.FC<{ path: string[]; rootKey?: string }> = ({
+  path,
+  rootKey,
+}) => {
+  const state = useFieldState(path);
+  const dispatch = useDocumentDispatch()!;
+
+  let childFields = null;
```
yuchenshi: nit: rename to `childEditors`.

src/components/Firestore/Field/index.tsx
```diff
@@ -14,37 +14,206 @@
...
+    </>
+  );
+};
+
+/**
+ * Field with CTAs for editing as well as rendering applicable child-nodes
+ */
+const NestedField: React.FC<{
+  path: string[];
+  edit?: boolean;
```
yuchenshi: where is this `edit` prop used?

src/components/Firestore/Field/index.tsx
```diff
@@ -14,37 +14,206 @@
...
+
+  function handleAddSave(value: FirestoreAny) {
+    dispatch(actions.addField({ path: addPath, value }));
+    setIsAddingField(false);
+  }
+
+  let addPath: string[] = [];
+  let childFields = null;
+  if (isMap(state)) {
+    addPath = [...path, ''];
```
yuchenshi: Is this `""` intentional? If so, why do we need this?

src/components/Firestore/Field/utils.ts
```diff
@@ -15,9 +15,18 @@
...
-import { FieldType } from './models';
 
-export function getFieldType(value: any): FieldType {
+import { FieldType, FirestoreAny } from './models';
+
+export function getParentPath(path: string[]) {
+  return path.splice(0, path.length - 1);
+}
+
+export function getLeafPath(path: string[]) {
```
yuchenshi: Consider renaming to `getLeafKey`? The returned value is a key (`string`) instead of path (`string[]`).

src/components/Firestore/Field/DocumentStore.tsx
```diff
@@ -0,0 +1,106 @@
+/**
```
yuchenshi: nit: consider renaming to `store.tsx`? We don't actually have a `DocumentStore` class.

src/components/Firestore/index.tsx
```diff
@@ -54,7 +56,24 @@ export const Firestore: React.FC<Props> = ({ config, projectId }) => {
   }
 
   async function handleClearData(api: DatabaseApi) {
-    const shouldNuke = await promptClearAll();
+    const shouldNuke = await confirm({
```
tstirrat: This dialogs change got reverted, please undo (use the external file)

src/components/Firestore/Field/DocumentStore.tsx
tjlav5: That third param only works for resolved undefined's... `The value returned for undefined resolved values.`. In this case, get does not resolve anything.

src/components/Firestore/Field/DocumentStore.tsx
tjlav5: It should never happen... we can assume and the types are defined such that only FirestoreArray and FirestoreMap can have child elements in the UI

src/components/Firestore/Field/DocumentStore.tsx
tjlav5: See above

src/components/Firestore/Field/DocumentStore.tsx
tjlav5: Undefined isn't a valid field-state. It has to return a type that the firebase sdk would be happy with, in this case I figure a string is our base type that we want to show users.

src/components/Firestore/Field/Editor.tsx
tjlav5: Updated, (Call to action)

src/components/Firestore/Field/index.tsx
tjlav5: Removed

src/components/Firestore/Field/Editor.tsx
tjlav5: Done

src/components/Firestore/Field/index.tsx
tjlav5: Yeah, added comment inline. This is the default key we pass into the child-editor.

src/components/Firestore/Field/DocumentStore.tsx
tjlav5: Done

src/components/Firestore/Field/utils.ts
tjlav5: Done

src/components/Firestore/index.tsx
tjlav5: Good catch, reverted

src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -90,19 +88,17 @@ const NestedEditor: React.FC<{ path: string[]; rootKey?: string }> = ({
...
-        <TextField
-          readOnly
-          value={getLeafPath(path) || rootKey}
-          placeholder="Field"
-        />
+        <TextField readOnly value={getLeafKey(path)} placeholder="Field" />
         <TextField readOnly value={getFieldType(state)} placeholder="Type" />
         <TextField
-          value={state}
+          value={JSON.stringify(state)}
```
yuchenshi: Why `stringify`? Is this going to turn the string `foo` into `"foo"` in the editor?

src/components/Firestore/DocumentPreview/api.tsx
```diff
@@ -0,0 +1,145 @@
...
+      //   foo: [  <-- foo == topLevelArray, nestedObjectToModify
+      //     'a',
+      //     'b',  <-- 1 == nestedFieldId (index, to be deleted)
+      //     'c',
+      //   ]
+      // }
+
+      // nestedObjectToModify.splice(Number(nestedFieldId), 1);
+      return {
+        [topLevelKey.join('.')]: firestore.FieldValue.arrayRemove(
```
yuchenshi: Corner case: `[1,2,1]` and the developer selects to delete the third item -- this will result in BOTH `1` to be removed.

src/components/Firestore/DocumentPreview/api.tsx
```diff
@@ -0,0 +1,145 @@
...
+    }
+    if (nestedObjectToModify instanceof Array && !topLevelArray) {
+      topLevelArray = nestedObjectToModify;
+      topLevelKey = fieldPath.slice(0, index + 1);
+    }
+  });
+
+  // If no array was found, carry on as usual
+  if (!topLevelArray) {
+    return { [fieldPath.join('.')]: value };
```
yuchenshi: Unfortunately Firestore fields can have `.` in them, so please use the `doc.update(new FieldPath(fieldPath), value)` API instead.

yuchenshi: Same for all joins below.

src/components/Firestore/DocumentEditor/index.tsx
tjlav5: ack this is going to be addressed in a follow up CL... this is just to get something printed for now

src/components/Firestore/DocumentPreview/api.tsx
tjlav5: Good catch, updated w/ tests

src/components/Firestore/DocumentPreview/api.tsx
tjlav5: Ah updated w/ tests

src/components/Firestore/utils.ts
```diff
@@ -15,9 +15,24 @@
...
+  FirestoreArray,
+  FirestoreMap,
+  FirestorePrimitive,
+} from './models';
+
+export function getParentPath(path: string[]) {
+  return path.slice(0, path.length - 1);
+}
+
+export function getLeafKey(path: string[]) {
```
tstirrat: FYI: this is lodash's `_.last(arr)` and the type can be `(input: T[]) => T`

supernit (totally optional): call it `last`?

src/components/Firestore/DocumentPreview/api.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Firestore does not allow array syntax when updating/deleting fields i.e.
+ * .update('obj.array[2]', true)
+ *
+ * This normalizes a payload's `objectPath` and `updateValue` into a form
+ * that writes to the top most array with a reconstructed payload contains the
+ * updated (or deleted) value.
+ */
+function adjustPayloadForArray(
```
tstirrat: This old friend lives on! I recall console having something very similar (is this the same code?)

tstirrat:
This is awesome btw. So much happening here and such a concise and clear PR. Very easy to grok.

src/components/Firestore/utils.ts
tjlav5: Cool swapped this out for lodash.last

src/components/Firestore/DocumentPreview/api.tsx
tjlav5: Yeah with some minor tweaks, but yeah same code... probably worth a second look in a follow-up

src/components/Firestore/DocumentEditor/store.tsx
```diff
@@ -0,0 +1,109 @@
...
+const reducer = createReducer<FirestoreMap, Action>({})
+  .handleAction(actions.reset, (state, { payload }) => {
+    return payload;
+  })
+  .handleAction(
+    actions.addField,
+    produce((draft, { payload }) => {
+      const parent = get(draft, getParentPath(payload.path)) || draft;
+      if (isMap(parent)) {
+        parent[last(payload.path) as string] = payload.value;
```
tstirrat: I think `last<string>()` is more safe, because it will check the type of the input array, vs cast the result.

PR #34, opened 2020-03-13
Style, grid tweak, add breadcrumbs

tstirrat:
* Restores bg canvas color
* Fixes rounded corners on cards
* Central grid 1256px wide (still 12 column)
* Add card action bar to each main card

Screen Shot 2020-03-13 at 1 40 30 PM
Screen Shot 2020-03-13 at 1 40 35 PM
Screen Shot 2020-03-13 at 1 40 41 PM

PR #36, opened 2020-03-14
Refactor database page into DatabaseContainer.

yuchenshi:
The new component will handle database list refreshing logic in a follow up PR. Tests are already in place though to show the desired behavior.

I've also fixed layout issues when `DatabasePicker` is shown, introduced by #34 .


src/components/Database/DatabasePicker.scss
```diff
@@ -24,11 +24,11 @@ a.mdc-list-item.mdc-list-item--activated {
 
 .Database-Picker {
   background-color: #fafafa;
-  border-radius: 16px 0 0 16px;
   border-right: solid 1px #dcdcdc;
 
   .mdc-list-group__subheader {
-    margin: 0 16px;
+    // Align horizontally with list icon and vertically with CardActionBar.
+    margin: 8px 16px 11px;
```
tstirrat: Generally we want multiples of 4px only, and at least one axis (in this case vertical) should usually use flex. Can this be flexed with only minimal alignment for icons?

Having said that, also happy for you to leave as is, and I will do a pass on alignment soon.

yuchenshi: This one cannot be flexed since it's just a paragraph, but let's definite fix the alignment later once we have the correct font size.

PR #37, opened 2020-03-14
Buttons/list bgs, dialog polish

tstirrat:
* I had to refactor material overrides. Dialogs and anything put into the top of the DOM would not take styles (i.e. buttons on dialogs were CAPITALIZED, vs not elsewhere), so I've moved it into `components/common/_material-overrides.scss` and ensured it loads after the material imports.
* I also made all the themes and primary/cancel buttons consistent (`unelevated` for non primary buttons).
* Fixed overflow issues on cards (remove .canvas-card)
* Fixed incorrect theme color in RTDB
* Dialog border radius is 16px

Screen Shot 2020-03-13 at 5 26 37 PM
Screen Shot 2020-03-13 at 5 26 47 PM


src/components/common/_material-overrides.scss
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import './components/common/radius';
+
+.mdc-card,
+.mdc-button {
+  border-radius: $BORDER_RADIUS_CARD;
```
tjlav5: You should be able to use

@include mdc-button-shape-radius($BORDER_RADIUS_CARD);

and

 @include mdc-card-shape-radius($BORDER_RADIUS_CARD);

src/components/common/_material-overrides.scss
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// These sometimes exist outside .App (e.g. DialogQueue)
+.mdc-dialog .mdc-dialog__surface {
+  border-radius: $BORDER_RADIUS_DIALOG;
```
tjlav5: Here too:

@include mdc-dialog-shape-radius($BORDER_RADIUS_DIALOG);

PR #38, opened 2020-03-14
VSCode: use workspace TS (fixes RMWC importing)

tstirrat:


yuchenshi:
LGTM, but do we have any idea why things are behaving like that? Is it the `typescript` version... or?

tstirrat:
Honestly I have no idea. I've tried to google it.. it just fails on the global version. I'll try `npm i -g typescript`

tstirrat:
Tried upgrading WS to 3.8.3 and it also broke. so this is some TS version issue :(

PR #43, opened 2020-03-17
(devx) Auto sort imports

tstirrat:
Uses `module` sorting. there are others, but this closely matches g3 style:

https://github.com/renke/import-sort/tree/master/packages/import-sort-style-module


src/components/Database/DataViewer/InlineQuery.tsx
```diff
@@ -14,26 +14,27 @@
...
-  QueryParams,
-  DEFAULT_PAGE_SIZE,
-  DEFAULT_QUERY_PARAMS,
-  jsonIshValue,
-} from './common/view_model';
-import { Select } from '@rmwc/select';
-import { TextField } from '@rmwc/textfield';
-import { useState, useEffect, useRef } from 'react';
-import keycode from 'keycode';
+import './InlineQuery.scss';
```
tjlav5: DId this get auto-moved to the top?

src/components/Database/Database.tsx
```diff
@@ -14,21 +14,23 @@
  * limitations under the License.
  */
 
-import React, { useState, useEffect, useCallback } from 'react';
-import * as firebase from 'firebase/app';
 import 'firebase/database';
```
tjlav5: Same question here, is this after sorting?

src/components/Database/DataViewer/InlineQuery.tsx
tstirrat: yep, side effect imports (no variable assignment) are put at top.

src/components/Database/Database.tsx
tstirrat: Yep, side effects go to the top, these are in the api.ts, so likely don't need to be here anyway

PR #44, opened 2020-03-17
Field editors for primitive types

tjlav5:
Need to follow-up with Maps and Arrays, as well as DocumentRef. Tests to come, but feel free to comment on direction.

### DIFF
--- a/src/components/Firestore/DocumentEditor/NumberEditor.tsx
+++ b/src/components/Firestore/DocumentEditor/NumberEditor.tsx
@@ -1,0 +1,50 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { useState, useEffect } from 'react';
+import { TextField } from '@rmwc/textfield';
+
+const NumberEditor: React.FC<{
+  value: number;
+  onChange: (value: number) => void;
+}> = ({ value, onChange }) => {
+  const [state, setState] = useState<string | number>(value);
+
+  useEffect(() => {
+    setState(value);
+  }, [value, setState]);
+
+  function handleChange(value: string) {
+    const newValue = parseInt(value);
+    if (!isNaN(newValue)) {
+      onChange(newValue);
+    } else {
+      setState(value);
+    }
+  }
+
+  return (
+    <TextField
+      label="Value"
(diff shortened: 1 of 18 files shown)


src/components/Firestore/utils.ts
```diff
@@ -66,6 +66,32 @@ export function getFieldType(value: FirestoreAny): FieldType {
   return typeof value.valueOf() as FieldType;
 }
 
+export function isBoolean(value: FirestoreAny): value is boolean {
+  return !!value && typeof value.valueOf() === 'boolean';
```
yuchenshi: `!!value` will be `false` when `value === false`. Probably not what you want.

src/components/Firestore/utils.ts
```diff
@@ -66,6 +66,32 @@ export function getFieldType(value: FirestoreAny): FieldType {
...
+): value is firestore.DocumentReference {
+  return value instanceof firestore.DocumentReference;
+}
+
+export function isTimestamp(value: FirestoreAny): value is firestore.Timestamp {
+  return value instanceof firestore.Timestamp;
+}
+
+export function isString(value: FirestoreAny): value is string {
+  return !!value && typeof value.valueOf() === 'string';
```
yuchenshi: `""`

src/components/Firestore/utils.ts
```diff
@@ -66,6 +66,32 @@ export function getFieldType(value: FirestoreAny): FieldType {
...
+}
+
+export function isNumber(value: FirestoreAny): value is number {
+  return !!value && typeof value.valueOf() === 'number';
```
yuchenshi: `0`

yuchenshi:
nit: One file per editor seems overkill but I don't feel strongly about it.

src/components/Firestore/DocumentEditor/NumberEditor.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onChange: (value: number) => void;
+}> = ({ value, onChange }) => {
+  const [state, setState] = useState<string | number>(value);
+
+  useEffect(() => {
+    setState(value);
+  }, [value, setState]);
+
+  function handleChange(value: string) {
+    const newValue = parseInt(value);
```
tstirrat: `parseInt` truncates decimals.

```js
Number('1234.5');
```

src/components/Firestore/utils.ts
tjlav5: Updated to using optional chaining

src/components/Firestore/utils.ts
tjlav5: Updated to using optional chaining

src/components/Firestore/utils.ts
tjlav5: Updated to using optional chaining

src/components/Firestore/DocumentEditor/NumberEditor.tsx
tjlav5: Good catch, switched to parseFloat... I'm not using Number since it returns 0 for non-parsed values.

yuchenshi: `parseFloat("1.2garbage")` will happily return `1.2` though. `Number` on the other hand gives you `NaN` for complete garbage or trailing garbage.

PR #45, opened 2020-03-17
Automatically fetch databases and show callout.

yuchenshi:
Tests pending for the new saga for polling. But other parts should be ready for review.

Screen Shot 2020-03-17 at 4 07 56 PM

### DIFF
--- a/src/colors.tsx
+++ b/src/colors.tsx
@@ -15,6 +15,7 @@
  */
 export const background = '#F6F7F9';
 export const primary = '#681DA8';
+export const secondary = '#1B3A57';
 
 export const grey100 = '#F5F5F5';
 
--- a/src/colors.tsx
+++ b/src/colors.tsx
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 export const background = '#F6F7F9';
-export const primary = '#681DA8';
-export const secondary = '#1B3A57';
+export const primary = '#681DA8'; // Google Purple 900.
+export const secondary = '#1B3A57'; // Firebase Navy 600.
 
 export const grey100 = '#F5F5F5';
 
-export const navBarPrimary = '#D7AEFB';
+export const navBarPrimary = '#D7AEFB'; // Google Purple 200.
 export const navBarOnSurface = '#FFF';
--- a/src/components/Database/DatabaseContainer.tsx
+++ b/src/components/Database/DatabaseContainer.tsx
@@ -1,11 +1,13 @@
+import { Button } from '@rmwc/button';
 import { Card } from '@rmwc/card';
 import { Elevation } from '@rmwc/elevation';
 import { GridCell } from '@rmwc/grid';
-import React, { ReactElement, useEffect } from 'react';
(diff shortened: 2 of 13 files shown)


src/colors.tsx
```diff
@@ -15,6 +15,7 @@
  */
 export const background = '#F6F7F9';
 export const primary = '#681DA8';
+export const secondary = '#1B3A57';
```
tstirrat: Can you add a comment with the Material color?

src/components/Database/DatabaseContainer.tsx
```diff
@@ -29,26 +32,62 @@ export const DatabaseContainer: React.FC<Props> = ({
...
+    return databasesUnsubscribe; // Unsubscribe when unmounting.
+  }, [databasesSubscribe, databasesUnsubscribe]);
+  const [dbs, setDbs] = useState(databases);
+
+  let hasNewDbs = false;
+  if (databases !== dbs) {
+    if (
+      dbs === undefined ||
+      databases === undefined ||
+      dbs.join('|') === databases.join('|')
```
tstirrat: Are they always sorted? if not this could fail

if so, maybe call the var sortedDbs

src/components/common/Callout.scss
```diff
@@ -0,0 +1,40 @@
+.Callout {
+  display: flex;
+  position: relative;
+
+  margin-bottom: 24px;
```
tstirrat: Don't make this mistake I made in console and regret every day. parent layout belongs in parent components. no padding or you'll have to keep removing and changing it elsewhere. :) remove this :)

src/store/database/sagas.ts
```diff
@@ -76,7 +85,23 @@ export function* fetchDatabases(
   }
 }
 
+const POLL_DATABASE_COOLDOWN_MS = 5000;
+
+export function* pollDatabases(): Generator<any> {
+  while (true) {
+    yield take(getType(databasesSubscribe));
+    while (
+      yield select((state: AppState) => state.database.databasesSubscribed)
```
tstirrat: Let's get into the habit of using `reselect` memoized selectors for all state queries. It's not totally important here, but we should build the patterns for future uses

src/colors.tsx
tstirrat: i.e. `// MATERIAL_GREY_200`

src/components/Database/DatabaseContainer.tsx
tstirrat: Or add a comment that they are sorted :)

src/colors.tsx
yuchenshi: Add a few comments. It actually took me a while to realize that this is a Firebase color, not a Material one.

src/components/common/Callout.scss
yuchenshi: Point taken.

src/store/database/sagas.ts
yuchenshi: I'll follow up with a separate PR to refactor all selectors we have so far.

PR #47, opened 2020-03-18
Callout colors and polish

tstirrat:
Screen Shot 2020-03-18 at 2 15 20 PM
Screen Shot 2020-03-18 at 2 16 05 PM
Screen Shot 2020-03-18 at 2 16 18 PM
Screen Shot 2020-03-18 at 2 16 26 PM
Screen Shot 2020-03-18 at 2 16 33 PM


src/components/common/Callout.scss
```diff
@@ -1,35 +1,127 @@
...
+  border-top-right-radius: $BORDER_RADIUS_CARD;
+}
+
+.mdc-card > .Callout:last-child {
+  border-bottom-left-radius: $BORDER_RADIUS_CARD;
+  border-bottom-right-radius: $BORDER_RADIUS_CARD;
+}
+
+// THEMES
+$msg-types: (
```
yuchenshi: Thoughts on using functions instead of list + loop?

tstirrat: If were going to re-asses this, lets keep as is. it is likely going to be deleted.

tstirrat:
Usage:

Screen Shot 2020-03-18 at 3 20 16 PM

tstirrat:
@tjlav5 we'll be switching to RMWC theming after Bug Bash.

And re tip, it should use RMWC theme's `--mdc-primary`. I didn't know the right way to wire that up... so I left it out because it's not needed yet.

Eventually this will consolidate around RMWC themes.

PR #48, opened 2020-03-18
Add indentation for nested data.

yuchenshi:
Screen Shot 2020-03-18 at 1 43 29 PM

### DIFF
--- a/src/components/Firestore/DocumentPreview/FieldPreview.scss
+++ b/src/components/Firestore/DocumentPreview/FieldPreview.scss
@@ -1,0 +1,49 @@
+$list-item-padding: 16px;
+$icon-size: 18px;
+$inline-spacing: 8px;
+
+.FieldPreview {
+  > .FieldPreview-expand-icon {
+    color: #757575;
+    vertical-align: middle;
+    margin-right: $inline-spacing;
+  }
+
+  &--primitive {
+    .FieldPreview-expand-icon {
+      visibility: hidden; // Hide it but keep it in layout to align with objs.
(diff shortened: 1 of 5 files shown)


src/components/Firestore/DocumentPreview/FieldPreview.scss
```diff
@@ -0,0 +1,49 @@
...
+    display: none;
+  }
+}
+
+.FieldPreview-children > .FieldPreview {
+  padding-left: ($list-item-padding + $icon-size + $inline-spacing);
+
+  .FieldPreview-children & {
+    padding-left: ($list-item-padding + ($icon-size + $inline-spacing) * 2);
+    .FieldPreview-children & {
```
tjlav5: This only goes 4 levels deep... should we be doing more? If we wanna go infinite... we could set a css-var on each FieldPreview that is the 'depth' of the field. Then the css here would be pretty simple:

--depth: {length of field}

padding-left: ($list-item-padding + ($icon-size + $inline-spacing) * var(--depth));

yuchenshi: Firebase Console only goes that deep. I don't doubt that we can keep going but at that point the UI barely usable due to the crazy indentation. We may revisit though.

Also, FYI max nesting is 20 levels, as publicly documented:  https://firebase.google.com/docs/firestore/quotas

tjlav5: Cool cool, good to know

PR #53, opened 2020-03-19
Conditionally mutable keys in firestore

tjlav5:
image

Keys for fields are mutable if they are new and not an array child. Tests to come in the morning...


src/components/Firestore/DocumentPreview/InlineEditor.tsx
```diff
@@ -42,15 +41,19 @@ const InlineEditor: React.FC<{
 
   function handleSave(e: React.MouseEvent<HTMLButtonElement>) {
     e.stopPropagation();
-    // TODO key could change in the editor
-    onSave(lastFieldName(path), internalValue);
+    console.log(internalValue);
```
yuchenshi: remove this

src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -71,8 +76,12 @@ const RootField: React.FC<{
 
   return (
     <>
-      {Object.keys(state).map(field => (
-        <NestedEditor key={field} path={[field]} />
+      {Object.keys(state).map((field, i) => (
```
yuchenshi: This one can be problematic especially now we key on index. When renaming fields, the iteration order for object keys can change and e.g. the current editor may end up moved to the last one.

I think we may need some special handling to keep things stable (e.g. local caching of existing keys)?

src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -81,20 +90,23 @@ const RootField: React.FC<{
...
+  path: string[];
+  isKeyMutable: boolean;
+}> = ({ path, isKeyMutable }) => {
   const state = useFieldState(path);
   const dispatch = useDocumentDispatch()!;
 
   let childEditors = null;
   if (isMap(state)) {
-    childEditors = Object.keys(state).map(childLeaf => {
+    childEditors = Object.keys(state).map((childLeaf, i) => {
```
yuchenshi: Same

src/components/Firestore/DocumentEditor/index.tsx
yuchenshi: Anyway, this is tricky, please change to `key={field}` and dispatch on blur for now (which is at least correct) and let's talk about how to optimize later.

src/components/Firestore/DocumentPreview/InlineEditor.tsx
tjlav5: Done

src/components/Firestore/DocumentEditor/index.tsx
tjlav5: Switched back to onBlur... it is still not perfect, but it'll do for EAP. There is an issue where when you enter in a field-name and then click on the type-selector or value-input, technically those elements are re-rendered when the blur-callback is triggered, thus requiring clicking on those elements a second time :(

tjlav5: I'll add local-caching as a p1 for next week

src/components/Firestore/DocumentEditor/index.tsx
tjlav5: Done

src/components/Firestore/DocumentEditor/store.tsx
```diff
@@ -64,6 +64,40 @@ const reducer = createReducer<FirestoreMap, Action>({})
...
+      const defaultValue = defaultValueForType(payload.type);
+      if (isMap(parent)) {
+        parent[lastFieldName(payload.path)] = defaultValue;
+      } else if (isArray(parent) && !isArray(defaultValue)) {
+        parent[Number(lastFieldName(payload.path))] = defaultValue;
+      }
+    })
+  )
+  .handleAction(
+    actions.updateKey,
```
yuchenshi: This one seems like a duplicate. Another identical handler for actions.updateKey above.

yuchenshi: Ah actually `updateType` is also a duplicate. Please double check the whole file.

PR #61, opened 2020-03-20
Callout colors and RMWC theme colors

tstirrat:
Colours:
Screen Shot 2020-03-20 at 10 56 44 AM

Buttons:
Screen Shot 2020-03-20 at 10 57 34 AM


src/index.tsx
```diff
@@ -52,7 +60,17 @@ ReactDOM.render(
...
     ripple={false}
   >
-    <ThemeProvider options={{ background, primary, secondary }}>
+    <ThemeProvider
+      options={{
+        background,
+        primary,
+        secondary,
+        error,
+        success,
```
yuchenshi: Please move `success`, `note`, `caution` out of the `--mdc-theme-` namespace and move them back to `--fire-color-`. `error` can stay since it overwrites material theme color, but `success` etc. won't play nicely with the rest of the ecosystem, for example `<Theme use="success">` won't pass type checking, and won't work even with `as any` since there is no underlying class for it. Similarly, nobody provides `successBg`, `onSuccess` etc., unlike `error`.

If applying `success` etc. to text ever become a common thing, we can roll out own helpers. Let's just stick to `var(--fire-color-success)` in CSS until then.

src/components/common/Callout.scss
```diff
@@ -55,73 +56,38 @@
...
+  }
+}
 
-      a,
-      a:hover,
-      .rmwc-icon,
-      .material-icons {
-        color: $dark;
-      }
+.Callout-warning {
```
yuchenshi: We should considering moving this to TSX and use `<ThemeProvider options={successTheme}/>` instead, in case we (re-)introduce more than one color for the success theme.

For example, I can imagine `const successTheme={primary: 'xxx', textPrimaryOnDark: 'xxx', ...}` if we ever want to use exact colors and variants like console.

src/components/common/Callout.scss
```diff
@@ -55,73 +56,38 @@
...
+    position: absolute;
+    right: 0;
+    top: 0;
+  }
 
-    // ASIDES
-    &.Callout-aside {
-      background-color: $light;
-      color: $dark;
+  > * {
```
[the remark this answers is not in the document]
tstirrat: oh, thats how!

src/index.tsx
tstirrat: sounds good :)

src/components/common/Callout.scss
tstirrat: yeah true.

src/components/common/Callout.scss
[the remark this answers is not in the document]
tstirrat: Obsolete, used background from theme

src/index.tsx
tstirrat: Done

src/components/common/Callout.scss
tstirrat: Done.

tstirrat:
Screens in original desctription updated.

PR #66, opened 2020-03-20
Use simpler algorithm for updating fields in arrays.

yuchenshi:
WANT_LGTM=all

This is complicated logic even if I tried to simplify it. Reviewers, please make sure it is easy to follow.


src/components/Firestore/DocumentPreview/api.tsx
```diff
@@ -55,104 +55,32 @@ function adjustPayloadForArray(
...
-    if (
-      topLevelArray === nestedObjectToModify &&
-      Number(lastFieldName(fieldPath)) >= (topLevelArray as {}[]).length
-    ) {
-      // Appending a new element to the top-level array
-      return [
-        new firestore.FieldPath(...topLevelKey),
-        firestore.FieldValue.arrayUnion(value),
-      ];
+  let cur: FirestoreAny | undefined = doc;
```
tjlav5: When would current be undefined?

tstirrat: perhaps on a missing doc when deep navigating?

tstirrat: Oh sorry, yes you're right. it doesn't look like it would be undefined ever

yuchenshi: `cur` will be undefined when adding a key to an object (e.g. set path `f.g` on `{"f": 1}`), with the assignment below. `doc` won't ever be `undefined` but `cur` will, thus the typing here.

PR #67, opened 2020-03-21
🐧 RTDB url routing and linking

tstirrat:
Much improved routing and linking for RTDB. Includes real href, so opening in new window works.

And it's fully router based now.

Cast: http://go/scrcast/NTQwNDUxMzA2ODM4NDI1NnxlOWU3Mjg1YS02Yg

### DIFF
--- a/src/components/Database/Database.tsx
+++ b/src/components/Database/Database.tsx
@@ -21,7 +21,7 @@ import { IconButton } from '@rmwc/icon-button';
 import { ListDivider } from '@rmwc/list';
 import { MenuItem, SimpleMenu } from '@rmwc/menu';
 import * as firebase from 'firebase/app';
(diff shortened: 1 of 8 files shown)


src/components/Database/Database.tsx
```diff
@@ -73,7 +71,7 @@ export const Database: React.FC<Props> = ({ config, namespace }) => {
         <>
           <InteractiveBreadCrumbBar
             base={urlBase}
-            path={new URL(ref.toString()).pathname}
+            path={path || '/'}
```
yuchenshi: Does `path` also need a trailing slash here?

tstirrat: No, trailing slashes are actually bad in the breadcrumbs. I will add tests in another PR. For now this gets deep linking working.

PR #73, opened 2020-03-24
Comprehensive button styles

tstirrat:
This is a little complex, but it's a port of the work in console done by the wonderful Roman N

http://go/firebase-spec/components/buttons

Screen Shot 2020-03-23 at 8 38 25 PM

Needs one more thing to be complete:

* Note:`theme="primary"` inside `"use={['primaryBg', 'onPrimary']}"` should be white (this is correct), but *any other custom theme on primary* should be the theme color i.e. the Warning button should be a big red button on purple. I'm leaving this as a TODO becauase those are less likely to happen, yet primary and secondary on primary/dark BG is more likely than warning on dark bg


src/components/common/_material-overrides-buttons.scss
```diff
@@ -0,0 +1,293 @@
...
+
+  @include __for-text-buttons-of-type(stroked, secondary) {
+    @extend %all-dark-hairline-buttons;
+  }
+
+  // Dark flat primary, secondary, and warn (all the same)
+  %all-dark-flat-buttons {
+    @include __flat-text-button-styling(
+      // sorry about important, overrides mdc important
+        $foreColor: $COLOR_WHITE_PRIMARY !important,
```
yuchenshi:
```suggestion
      $foreColor: $COLOR_WHITE_PRIMARY !important,
```

tstirrat:
> Code LGTM, but can we look into [Storybook or Styleguidist](https://create-react-app.dev/docs/developing-components-in-isolation/#getting-started-with-storybook) to test and showcase our components? An additional route just for style guide bloats our distribution size.

Yeah, that sounds good. I'll add a TODO about that caveat and remove the styleguilde routes/page

src/components/common/_material-overrides-buttons.scss
tstirrat: It's auto formatted, so it reverts. will leave :)

tstirrat:
Colored on primary is fixed by forcing `theme="primary"` when inside a `theme="primaryBg onPrimary"`. The focus states are working but dimm (due to opacity), I will punt on this.

Screen Shot 2020-03-24 at 1 03 42 PM

PR #83, opened 2020-03-26
Refactor store and handle loading / error in tabs.

yuchenshi:
This PR has a large diff, but it's mostly a few global changes to support loading / error states with little visual impact.

1. The store has been refactored to use the unified `RemoteResult<T>` for all sorts of things that can be fetched from online.
   * In short it's just `{loading: boolean, result: {data: T}}` (or `result: {error: E}` in case of error).
   * Start from `src/store/util.ts` which introduces the type alias and the helper functions
   * The shape is now flexible enough to allow refreshing (and polling) everywhere.
2. All selectors are now in `src/store/*/selector.ts`, decoupling store shape from components.
   * reselect is introduced with memorization built in.
   * A few selectors return `RemoteResult<T>` so that we can separately handle success, error, and missing cases in a type-safe manner.
3. All three routes now have a zero and error state components. Visual is TBD and will be done in another PR, to keep the PR size reasonable.
   * A global "disconnected" modal will land in a follow-up PR.
   * Note: the cases where each individual emulator is killed but GUI server still lives on is not handled since this is not supported yet.
   * For now, there are some boilerplate code across tabs but I'll leave it there for flexibility should the UX change. I'll come with with a better abstraction once it is finalized.

### DIFF
--- a/src/components/Firestore/index.tsx
+++ b/src/components/Firestore/index.tsx
@@ -24,8 +24,10 @@ import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
 import { useHistory, useLocation } from 'react-router-dom';
 
-import { AppState } from '../../store';
+import { createStructuredSelector } from '../../store';
 import { FirestoreConfig } from '../../store/config';
+import { getFirestoreConfig, getProjectId } from '../../store/config/selectors';
+import { combineData, squash } from '../../store/utils';
 import { CustomThemeProvider } from '../../themes';
 import { InteractiveBreadCrumbBar } from '../common/InteractiveBreadCrumbBar';
 import DatabaseApi from './api';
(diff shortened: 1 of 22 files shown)


src/components/Firestore/index.tsx
```diff
@@ -61,7 +84,7 @@ export const Firestore: React.FC<Props> = ({ config, projectId }) => {
   }, [projectId, config, setApi]);
 
   if (!api) {
-    return <p>Connecting to Firestore...</p>;
+    return <></>;
```
tjlav5: nit: just return null?

src/store/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+export type ErrorResult<E> = { error: E };
+
+export type RemoteResult<T, E = ErrorInfo> = Remote<Result<T, E>>;
+
+const ownProp = Object.prototype.hasOwnProperty;
+
+export function hasData<T, E>(
+  result: Result<T, E> | undefined
+): result is DataResult<T> {
+  return result !== undefined && ownProp.call(result, 'data');
```
tjlav5: I think you want to use 'in' here and below, it's a little more succinct

https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-the-in-operator

src/store/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+  return {
+    loading: remoteResult1.loading || remoteResult2.loading,
+    result:
+      remoteResult1.result === undefined || remoteResult2.result === undefined
+        ? undefined
+        : hasError(remoteResult1.result)
+        ? remoteResult1.result
+        : hasError(remoteResult2.result)
+        ? remoteResult2.result
+        : { data: [remoteResult1.result.data, remoteResult2.result.data] },
```
tjlav5: Can we stay away from chained ternaries? :)

src/components/Firestore/index.tsx
yuchenshi: Done.

src/store/utils.ts
yuchenshi: Updated. People back in the days tell me not to use `in` because it walks through the whole prototype chain, but whatever.

src/store/utils.ts
yuchenshi: I've changed this function to accept arrays with any length with an iteration-based implementation. TypeScript mapped tuples FTW!

PR #87, opened 2020-03-27
Better UX for null RTDB roots

tstirrat:
Screen Shot 2020-03-26 at 5 11 33 PM
Screen Shot 2020-03-26 at 5 20 51 PM

### DIFF
--- a/src/components/Database/DataViewer/ValueDisplay.tsx
+++ b/src/components/Database/DataViewer/ValueDisplay.tsx
@@ -37,7 +37,7 @@ export const ValueDisplay = React.memo<Props>(function ValueDisplay$({
     `ValueDisplay ValueDisplay__${typeof value}`;
   return (
     <Typography use="body1" className={getClass(value)} {...otherProps}>
-      {JSON.stringify(trimmedValue)}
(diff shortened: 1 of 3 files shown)


src/components/Database/DataViewer/ValueDisplay.tsx
```diff
@@ -37,7 +37,7 @@ export const ValueDisplay = React.memo<Props>(function ValueDisplay$({
     `ValueDisplay ValueDisplay__${typeof value}`;
   return (
     <Typography use="body1" className={getClass(value)} {...otherProps}>
-      {JSON.stringify(trimmedValue)}
+      {value === null ? '{}' : JSON.stringify(trimmedValue)}
```
yuchenshi: Let's still show it as `null`, since that's what you'll get back from SDK and REST apis. I don't want to see a bug report highlighting the disparency.

yuchenshi: To elaborate, the way we usually document and explain this goes: There is no actual `{}` in RTDB, only `null`. There's an RTDB feature that creates an object for you if you add a field to `null`, which is working as intended. And if you delete all fields in an object, it turns back into `null`.

PR #88, opened 2020-03-27
Use Firestore inline editor for RTDB

tstirrat:
Screen Shot 2020-03-26 at 6 10 19 PM
Screen Shot 2020-03-26 at 6 10 25 PM


tstirrat:
> LGTM, but let's definitely disable timestamps and other weird data types. For example, the RTDB SDK won't handle Firestore Timestamp or DocumentReference well.

Done:

Screen Shot 2020-03-27 at 12 45 22 PM

PR #91, opened 2020-03-28
Firestore validation

tjlav5:
image

Few things happening here:
1) Normalized document-state - handling name/type/value changes are trivial now
2) Validation... we're probably shoehorning react-hook-form into this at at this point. It really is a nice lib, but it does not do so well with Controlled components. The document-editor inputs are now all uncontrolled (default-value). This does help with performance, but it does complicate the data-store a little bit. Now there are essentially two sources of truth. There is the local ref, which has what the user inputted; be it valid or invalid. Then there is the document-store which is a strongly-typed representation of the last valid state.

Extra notes on validation:
* If the Editor is in an invalid state, it will emit undefined to anyone listening... this is how we can disable the various save buttons in the UI that exist outside of the editor.
* Each value editor has its own validation rules where applicable.
* The Field-input has a special validation rule that ensures we don't have redundant sibling key-names. It occurred to me that we will need to add an extra input prop to the Editor {non-editable root-keys}; e.g. if you are inline-adding an entry to a pre-existing map, then the editor is actually not aware of those sibling keys but we should still err if they re-use a key.
* Testing honestly started to get weird with... I had to add a bunch of async acts per the react-hook-form docs.

The UI is still rough, will need some help with the layout. The typings need to be beefed up a bit more too, I think we can make this better.

Lastly, there is one regression here, but it isn't an issue YET... in the process of normalizing the data, I have not added map/array support yet. I think it'll be a fun thing to add, but right now Editors are only every initialized with primitives in the root so we are OK for now......


src/components/Firestore/dialogs/AddCollectionDialog.test.tsx
```diff
@@ -125,9 +125,28 @@ describe('step 2', () => {
     expect(getByLabelText('Field')).not.toBeNull();
   });
 
+  // TODO testing suggests that the button is infact disabed by inspecting the DOM
+  // but triggering a click event still triggers the underlying event. This is no
+  // reproducible in the actual GUI.
+  it.skip('[Save] is disabled if invalid doc-data', async () => {
```
tstirrat: Totally can punt this later, but try wrapping dialog in a form (we want this later for nice UX on pressing enter), making buttons type="button" and type="submit",  then doing:

```js
fireEvent.submit(getByText('Save'))
```

This worked for me in clone dialog, and it *may* be smart enough to not submit on a disabled button.

tjlav5: Ack, will add this for a quick follow-up tomorrow. Adding to the b/ list

PR #112, opened 2020-03-30
Document-editor v2

tjlav5:
Re-thought the document-store over the weekend, this would allow us to actually edit data-sets that are not maps... this is going to come in handy in a quick follow-up to better handle adding an entry to an array, as opposed to faking a map with key equal to the array index. Also, this could help in the future if we want to beef up support in RTDB where the root node is not necessarily a JSON-like obj.

Note this doesn't really change the public API of document-editor, thus the tests are pretty much the same... might beef these up in the future, but should be good for now.

The new structure is better typed and decouples FieldType in DocumentPreview from Fields in DocumentEditor. General structure:
```
{
  rootId: _id_,
  fields: { id: MapField|ArrayField|PrimitiveField }
}
```

Components: `<DocumentEditor />`, `<FieldEditor />`. Dropped the `<RootEditor />` and merged functionality with core-DocumentEditor. Mitigated all re-render issues, super memoized and performant now. This will also help tremendously with Field-Input layouts since we can have properly nested trees.


src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -85,272 +88,300 @@ export function supportsEditing(value: FirestoreAny): boolean {
...
-      if (onChange) {
-        if (errorCount === 0) {
-          onChange(sdkMap);
-        } else {
-          onChange(undefined);
-        }
-      }
-    }
+const FieldEditor: React.FC<{
+  id: number;
```
tstirrat: Can we make this id `uuid` or `guid` throughout, so that it's clear it's **not** the key of the field?

tjlav5: Good idea, replaced with UUID

PR #114, opened 2020-03-30
Implement general disconnected overlay.

yuchenshi:
Screen Shot 2020-03-30 at 2 08 25 PM


src/components/Home/index.test.tsx
```diff
@@ -66,7 +66,8 @@ it('shows port for emulator that are loaded', () => {
   expect(getByText(/9000/)).not.toBeNull();
 });
 
-it('renders error message when errored', () => {
+// Updated error screen is WIP for Home for now.
+it.skip('renders error message when errored', () => {
```
tstirrat: remove .skip

yuchenshi: This is intentionally skipped (based on the comment above) and I'll redo the test in the next PR.

tstirrat: Oh, I see :) maybe put `// intentional skip:` or something like that.

PR #118, opened 2020-03-31
Use immer for reducers and reuse exisiting objects.

yuchenshi:
Removes the crazy rerendering for Firestore + RTDB.


src/store/config/reducer.ts
```diff
@@ -14,18 +14,30 @@
  * limitations under the License.
  */
 
+import produce, { isDraft, original } from 'immer';
```
tjlav5:
```suggestion
import produce from 'immer';
```

src/store/database/reducer.ts
```diff
@@ -14,8 +14,11 @@
  * limitations under the License.
  */
 
+import produce from 'immer';
+import deepEquals from 'lodash.isequal';
```
tjlav5:
```suggestion
```

yuchenshi:
@tjlav5 I'm aware of this and I'll look into this. Merging this PR though to get GUI back to a usable (testable) state for now.

yuchenshi:
@tjlav5 Fix coming in #121

PR #128, opened 2020-04-01
Add loading state while clearing-firestore

tjlav5:


src/components/Firestore/index.tsx
```diff
@@ -94,14 +96,32 @@ export const Firestore: React.FC<FirestoreProps> = ({ config, projectId }) => {
   async function handleClearData(api: DatabaseApi) {
     const shouldNuke = await promptClearAll();
     if (!shouldNuke) return;
-    api.nukeDocuments();
+    setIsRefreshing(true);
+    api.nukeDocuments().then(() => {
```
tstirrat, line 100: no await inside async func? is that intentional?

src/components/Firestore/index.test.tsx
```diff
@@ -91,17 +94,27 @@ describe('Firestore', () => {
 
   it('triggers clearing all data', async () => {
     confirm.mockResolvedValueOnce(true);
-    const nukeSpy = jest.fn();
+    let nukeResolve;
+    const nukeSpy = jest
+      .fn()
+      .mockImplementation(
+        () => new Promise(resolve => (nukeResolve = resolve))
```
yuchenshi, line 101: Please also extract this `Promise` out to alongside nukeResolve so we can `await nukeResolve` instead of `await wait()` later.

src/components/Firestore/index.tsx
tjlav5: Ah of course, added.

PR #131, opened 2020-04-01
Deduplicate logs.

yuchenshi:
This PR contains two changes to fix duplicate logs.

1. Remove duplicate log history when navigating to Logs tab, navigate away, and then navigate back. This is because the WebSocket connection always send over all existing logs even on reconnect.
2. Fix cleanup for the socket connection to prevent memory leak and also removes duplicate listeners. This fixes an issue where the same new log will appear more than one time.

The fixes here are temporary to get the GUI into a better state while we're waiting for pending refactors to move websockets into sagas.

### DIFF
--- a/src/store/logviewer/actions.ts
+++ b/src/store/logviewer/actions.ts
@@ -19,3 +19,6 @@ import { createAction } from 'typesafe-actions';
 import { LogEntry } from './types';
 
 export const logReceived = createAction('@log/RECEIVED')<LogEntry>();
+
+// TODO: Remove this hack once it's not needed.
+export const logReset = createAction('@log/REST')();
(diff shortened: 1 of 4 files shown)


src/store/logviewer/actions.ts
```diff
@@ -19,3 +19,6 @@ import { createAction } from 'typesafe-actions';
 import { LogEntry } from './types';
 
 export const logReceived = createAction('@log/RECEIVED')<LogEntry>();
+
+// TODO: Remove this hack once it's not needed.
+export const logReset = createAction('@log/REST')();
```
tjlav5, line 24: s/REST/RESET

yuchenshi: Done!

PR #134, opened 2020-04-01
Add loading state for routes.

yuchenshi:
Screen Shot 2020-04-01 at 11 12 16 AM

### DIFF
--- a/src/components/Firestore/index.tsx
+++ b/src/components/Firestore/index.tsx
@@ -21,6 +21,7 @@ import { Card } from '@rmwc/card';
 import { CircularProgress } from '@rmwc/circular-progress';
 import { Elevation } from '@rmwc/elevation';
 import { GridCell } from '@rmwc/grid';
+import { Typography } from '@rmwc/typography';
 import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
 import { useHistory, useLocation } from 'react-router-dom';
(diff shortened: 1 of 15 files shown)


src/components/Firestore/index.tsx
```diff
@@ -147,9 +142,17 @@ export const Firestore: React.FC<FirestoreProps> = ({ config, projectId }) => {
...
+export const FirestoreLoading: React.FC = () => (
+  <GridCell
+    span={12}
+    className="Firestore-loading"
+    data-testid="firestore-loading"
+  >
+    <CircularProgress size="xlarge" />
+    <Typography use="body2" tag="p">
+      Firestore Emulator Loading...
+    </Typography>
```
tstirrat, line 154: This and the centering styles are reused, can you extract into a component?

In console we have `<fire-spinner message="Blah loading" />`, where the default is to show a spinner only. or include the message if provided.

I think something like that would be helpful.

tstirrat: And we could use without the grid and centering stuff with `<Spinner inline size="xsmall" />`

Perhaps you could even include the Grid and span={}.. its often going to be included as a grid flow: `<Spinner span={12} />`

tstirrat: The extra component stuff is easy to do later, but making it into a simple component `<Spinner message="test" />` is helpful now.

yuchenshi: Added the grid version now but I'll leave the `inline` version unimplemented until we actually need it.

src/components/Firestore/index.tsx
```diff
@@ -147,9 +143,17 @@ export const Firestore: React.FC<FirestoreProps> = ({ config, projectId }) => {
   );
 };
 
-export const FirestoreRouteLoading: React.FC = () => (
-  // TODO
-  <div>Loading...</div>
+export const FirestoreLoading: React.FC = () => (
```
tstirrat, line 146: is this dead code now?

PR #157, opened 2020-04-06
Disable auto merge for now.

yuchenshi:
I think googlebot and automerge workflow has triggered a loop with each other, spamming our Github Actions queue quickly. When this happens, normal CI workflows will wait very long (or even indefinitely) in the queue.

Therefore, I'm disabling automerge for now.


.github/workflows/automerge.yml
```diff
@@ -1,36 +0,0 @@
...
-      - edited
-      - ready_for_review
-      - reopened
-      - unlocked
-  pull_request_review:
-    types:
-      - submitted
-  check_suite:
-    types:
-      - completed
```
tstirrat, line 18: We might be able to remove this, so that it does not trigger on other checks?

yuchenshi: I guess automerge just won't run when CI complete if we do that, making it useless.

PR #158, opened 2020-04-06
Remove utils/uuid

tstirrat:
Missed comment from #151


tstirrat:
I'm going to override it in test with jest mocks :)

tstirrat:
Woohoo, already there: https://github.com/firebase/firebase-tools-ui/blob/93c19cf5b44f3b81e925846d9ad59392324cd48e/src/setupTests.js#L15-L16

PR #159, opened 2020-04-07
Remove await wait() in favor of better approaches.

yuchenshi:
WANT_LGTM=all

This PR is a massive refactor to remove `await wait()` in our tests in favor of better ways to test.

From now on, please refrain from `await wait()` at all cost in the tests, for the following reasons:

* `await wait()` is deprecated in the [next major release](https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0) of `dom-testing-library`.
* `await wait()` is fragile, since it relies implicitly on the next DOM update / tick. Sometimes it even takes more than one `await wait()`, leading to unreadable and undebuggable code. Other times, they make tests flaky.
* `await wait()` is not close to how users interact with the app.
    - Users do not perceive DOM implementation details, instead they perceive things like dialogs opening and closing (through monitors, screen readers, etc.).
    - See the [release notes of dom-testing-library](https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0) for more rationale and the goal of the whole library.

Instead:
- Use utils in `test_utils.ts` for the specific use cases they cover like Promises and dialogs.
- Use `wait(() => expect(...))` for assertions that eventually should succeed.
- Use `findByXXX` to get elements when they eventually appear in DOM.
    * Or use `waitForElement` if there is no `findByXXX` for that (e.g. class names). (Less ideal.)

As a last resort, we can use `await delay(200)` inside `act()`. It conveys the expectation that the component has complicated initial work / updates that finishes in 200ms, and implicitly asserts that user perception during that time does not matter since they are not expected to interact with the component. In my opinion, this is less ideal than waiting for more direct changes, but still better than `await wait()`.

### DIFF
--- a/src/components/Firestore/CollectionList.test.tsx
+++ b/src/components/Firestore/CollectionList.test.tsx
@@ -14,14 +14,16 @@
  * limitations under the License.
  */
 
-import { render, wait } from '@testing-library/react';
+import { act, render, wait } from '@testing-library/react';
 import React from 'react';
 import { MemoryRouter } from 'react-router-dom';
 
+import { makeDeferred, renderAndWait } from '../../test_utils';
 import DatabaseApi from './api';
 import { ApiProvider } from './ApiContext';
 import CollectionList from './CollectionList';
 import {
+  FakeCollectionReference,
   fakeCollectionReference,
   fakeDocumentReference,
   fakeDocumentSnapshot,
@@ -32,12 +34,10 @@ jest.mock('./api');
 
 it('shows the list of collections', async () => {
   const fakeApi = new DatabaseApi();
-  fakeApi.getCollections.mockResolvedValueOnce([
-    fakeCollectionReference({ id: 'coll-1' }),
-    fakeCollectionReference({ id: 'coll-2' }),
-  ]);
+  const getCollections = makeDeferred<FakeCollectionReference[]>();
+  fakeApi.getCollections.mockReturnValueOnce(getCollections.promise);
 
-  const { getByText, queryByText } = render(
+  const { getByText } = await render(
     <MemoryRouter>
       <ApiProvider value={fakeApi}>
         <CollectionList />
@@ -45,8 +45,12 @@ it('shows the list of collections', async () => {
     </MemoryRouter>
   );
 
-  await wait();
-
+  await act(() =>
+    getCollections.resolve([
+      fakeCollectionReference({ id: 'coll-1' }),
+      fakeCollectionReference({ id: 'coll-2' }),
+    ])
+  );
   expect(getByText(/coll-1/)).not.toBeNull();
   expect(getByText(/coll-2/)).not.toBeNull();
 });
@@ -54,15 +58,13 @@ it('shows the list of collections', async () => {
 it('requests root-collections with no docRef', async () => {
   const fakeApi = new DatabaseApi();
 
-  const { getByText, queryByText } = render(
+  await renderAndWait(
     <MemoryRouter>
       <ApiProvider value={fakeApi}>
         <CollectionList />
       </ApiProvider>
     </MemoryRouter>
   );
-
-  await wait();
 
   expect(fakeApi.getCollections).toHaveBeenCalledWith(undefined);
 });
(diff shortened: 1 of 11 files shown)


src/components/Firestore/index.test.tsx
```diff
@@ -77,29 +81,32 @@ describe('FirestoreRoute', () => {
...
-    const { getByText, queryByText } = render(
-      <MemoryRouter>
-        <Firestore config={sampleConfig} projectId={'foo'} />
-      </MemoryRouter>
+    let getCollections = makeDeferred<FakeCollectionReference[]>();
+    DatabaseApi.prototype.getCollections.mockReturnValue(
+      getCollections.promise
     );
+    let renderResult!: RenderResult;
+    act(() => {
```
tstirrat, line 89: afaik render already is wrapped in an act:

https://github.com/testing-library/react-testing-library/blob/master/src/pure.js#L53-L59

src/test_utils.ts
```diff
@@ -36,7 +44,115 @@ export async function renderAndWait(
   await act(async () => {
     renderResult = render(ui, options);
     // Silence warnings caused by async DOM updates.
-    await delay(100);
+    await delay(200);
```
tstirrat, line 47: might be able to use `jest.runAllTimers()` here (for another PR), I'm concerned that we'll need to keep pushing this up to fit the slowest render, slowing down all other tests.

https://jestjs.io/docs/en/timer-mocks

src/test_utils.ts
```diff
@@ -19,15 +19,23 @@ import {
...
 } from '@testing-library/react';
 
 export function delay(timeoutMs: number): Promise<void> {
   return new Promise(resolve => setTimeout(resolve, timeoutMs));
 }
 
-// Render a component and wait for some async DOM changes on init.
-// This can be used to silence warnings of changing the DOM outside
-// act(...), e.g. for RMWC <Tab>s or <Dialog>s.
+/**
```
tstirrat, line 31: Suggestion on name: `renderWithAsyncEffects(<div />)`

src/components/Firestore/index.test.tsx
yuchenshi: Done.

src/test_utils.ts
yuchenshi: Removed this function. See discussion above.

src/components/Firestore/CollectionList.test.tsx
```diff
@@ -32,37 +34,39 @@ jest.mock('./api');
...
   const fakeApi = new DatabaseApi();
-  fakeApi.getCollections.mockResolvedValueOnce([
-    fakeCollectionReference({ id: 'coll-1' }),
-    fakeCollectionReference({ id: 'coll-2' }),
-  ]);
+  const getCollections = makeDeferred<FakeCollectionReference[]>();
+  fakeApi.getCollections.mockReturnValueOnce(getCollections.promise);
 
-  const { getByText, queryByText } = render(
+  const { getByText } = await render(
```
tstirrat, line 40: I think this should just be `= render(` ?

src/components/Firestore/useAutoSelect.test.tsx
```diff
@@ -140,13 +149,13 @@ describe('useAutoSelect', () => {
           </Route>
         </Router>
       );
-      await wait();
+      await act(() => delay(100));
```
tstirrat, line 152: curious if these delays waiting on promises are fixable with `await act(() => jest.runAllTimers())` or similar?

else, LGTM!

src/test_utils.ts
tstirrat: 👍 love your work

src/components/Firestore/useAutoSelect.test.tsx
yuchenshi: `useAutoSelect` does not wait on promises, instead the problem is that it triggers `useEffect` which triggers `<Redirect>` which in turn re-renders. So I don't think `jest.runAllTimers()` can help here. Also there are many test cases that cover cases that should not trigger redirects -- just calling `jest.runAllTimers()` is not a great proof that it won't redirect in the next frame. I think the way to go is to mock out `<Redirect>` so it renders some text instead of side-effects, and I'm happy to review a PR for that.

For other promise-based stuff I've switched most to use Deferred. Deferred is the most reliable and controllable way to test those cases and we should prefer that to `jest.runAllImmediates()` if possible. (If a component uses Promise in a way that makes using Deferred hard, we should consider refactoring first, or turn fake timers on and call `jest.runAllImmediates()` just for a few tests.)

src/components/Firestore/CollectionList.test.tsx
yuchenshi: Oops. Done.

PR #160, opened 2020-04-08
Update to jsdom 16, remove mutationobserver shim

tstirrat:
jsDOM 16 Has more DOM emulation features


tstirrat:
Strange. unit tests seemed to work fine, including coverage on local

yuchenshi:
`BigInt` is not defined? What did they pull in in jsdom 16?

yuchenshi:
So the CI just failed with a different error message. If anything, it looks like there is a race condition in the test setup to me. Feel free to investigate if you have more time, or just hold off on jsdom 16 for now.

tstirrat:
Sadly, I don't really want to fix node 8 BigInt with another shim. its not exactly any worse than the mutation observer shim. will leave as is

PR #163, opened 2020-04-09
Improve node CI install and init speed

tstirrat:


yuchenshi:
Any comparison of before and after? From my casual observation, it seems Node.js 8 build actually became slower, increasing from the usual ~2m to [3m now](https://github.com/firebase/firebase-tools-ui/pull/163/checks?check_run_id=572565726). I could be wrong though.

Also, the package-lock check was broken.

tstirrat:
You're right. I got it down to 2 mins, but it seems npm i even with cache still requires 30s :/

The other repo was using yarn and it;s way faser on locked repos.

This does not look like enough benefit to be worth it

yuchenshi:
Actually, do you want to invest in switching to yarn then? It also gives us nice benefits like manually specifying version resolutions, allowing us to fix some of the pending security warnings easily.

PR #165, opened 2020-04-10
Tl/firestore nested arrays

tjlav5:
image

Better array support in the document editor.

1) conditionally allow nested-arrays (in the case of Firestore, this is false)
2) add a `startingIndex` to the editor, so that we get the look+feel of arrays when adding a new element, but getting the correct number to render in the index-input-field

### DIFF
--- a/src/components/Firestore/DocumentEditor/index.tsx
+++ b/src/components/Firestore/DocumentEditor/index.tsx
@@ -93,12 +93,14 @@ const DocumentEditor: React.FC<{
   areRootNamesMutable?: boolean;
   areRootFieldsMutable?: boolean;
   rtdb?: boolean;
+  startingIndex?: number;
 }> = ({
   value,
   onChange,
   areRootNamesMutable,
   areRootFieldsMutable,
   rtdb = false,
+  startingIndex,
 }) => {
   const initialState = normalize(value);
   const [store, dispatch] = React.useReducer(storeReducer, initialState);
@@ -134,6 +136,7 @@ const DocumentEditor: React.FC<{
                   isRtdb={rtdb}
                   areNamesMutable={areRootNamesMutable}
                   areFieldsMutable={areRootFieldsMutable}
+                  startingIndex={startingIndex}
                 />
               )}
             </>
@@ -152,7 +155,14 @@ const FieldEditor: React.FC<{
   isRtdb: boolean;
   areNamesMutable?: boolean;
   areFieldsMutable?: boolean;
-}> = ({ uuid, isRtdb, areNamesMutable = true, areFieldsMutable = true }) => {
+  startingIndex?: number;
+}> = ({
+  uuid,
+  isRtdb,
+  areNamesMutable = true,
+  areFieldsMutable = true,
+  startingIndex = 0,
+}) => {
   const store = useStore();
   const dispatch = useDispatch();
   const field = useField(uuid);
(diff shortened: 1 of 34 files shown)


src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -302,15 +325,20 @@ const FieldEditor: React.FC<{
 const ChildTypeSelect: React.FC<{
   uuid: number;
   isRtdb: boolean;
-}> = ({ uuid, isRtdb }) => {
+  blacklist?: FieldType[];
```
yuchenshi, line 328: Consider adding a `allowedTypes` prop to replace both `isRtdb` and `blacklist`. 
(Alternatively but less preferred, rename to `disabledTypes` for clarity.)

tstirrat: yeah, slight pref for a white list and remove rtdb, vs blacklist ..

src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -217,10 +232,18 @@ const FieldEditor: React.FC<{
...
+                      label="Index"
+                      disabled
+                    />
                   </div>
                   <span className="Document-TypeSymbol">=</span>
-                  <ChildTypeSelect uuid={c.valueId} isRtdb={isRtdb} />
+                  <ChildTypeSelect
+                    uuid={c.valueId}
+                    isRtdb={isRtdb}
+                    blacklist={!supportNestedArrays ? [FieldType.ARRAY] : []}
```
tstirrat, line 245: use const arrays since these will never change, and it will hopefully prevent a mysterious re-render later.

src/components/Firestore/DocumentEditor/index.tsx
tstirrat: Oh, I see how it's used. blacklist is more lightweight.. carry on :)

tstirrat: But I do think it's worth combining blacklist and isRtdb

src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -302,15 +325,20 @@ const FieldEditor: React.FC<{
 const ChildTypeSelect: React.FC<{
   uuid: number;
   isRtdb: boolean;
-}> = ({ uuid, isRtdb }) => {
+  blacklist?: FieldType[];
+}> = ({ uuid, isRtdb, blacklist = [] }) => {
   const field = useField(uuid);
   const dispatch = useDispatch();
 
+  const options = (isRtdb ? RTDB_FIELD_TYPES : FIRESTORE_FIELD_TYPES).filter(
```
tstirrat, line 333: Can this be memo'd for less re-renders? it'll only enumerate to 3 or 4 possibilities ever

src/components/Firestore/DocumentEditor/index.tsx
tjlav5: Ack, updated to allowedTypes and merged the two props.

src/components/Firestore/DocumentEditor/index.tsx
tjlav5: Done

src/components/Firestore/DocumentEditor/index.tsx
tjlav5: Pulled these out, and now just storing the "ARRAY_FIELD_TYPES" in a const outside of the component.

PR #171, opened 2020-04-14
Default to string entry when editing RTDB root

tstirrat:
Screen Shot 2020-04-14 at 12 42 05 PM

before:

Screen Shot 2020-04-14 at 1 09 54 PM

after:

Screen Shot 2020-04-14 at 12 42 10 PM


tstirrat:
For now that's a little wonky.. because maps need another click to add children.

once I update the maps editor to auto add a sub field.. then we can revisit it here

PR #173, opened 2020-04-14
RTDB -> Realtime Database

tstirrat:
I wonder if we can use Realtime DB for some labels?

Screen Shot 2020-04-14 at 4 50 19 PM

### DIFF
--- a/src/components/Home/index.scss
+++ b/src/components/Home/index.scss
@@ -4,17 +4,18 @@
   }
 
   .title {
-    margin-top: 8px;
+    align-items: flex-start;
     display: flex;
-    align-items: center;
+    margin-top: 8px;
+    min-height: 64px;
 
     .rmwc-icon {
       margin-right: 8px;
+      margin-top: 6px;
     }
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/components/Home/index.scss
```diff
@@ -16,4 +16,5 @@
 
 .Home-EmulatorCard-Info {
   margin: 16px;
+  min-height: 214px; // fix wrapping text making cards different sizes
```
yuchenshi, line 19: Let's just make the title of each card 2 line-heights tall. This should remove the need to calculate here and also hopefully aligns better.

tstirrat: I tried that and it was wonky on normal modes.. too much whitespace ...  for smaller screens, I went with ellipsis for now:

Screen Shot 2020-04-14 at 5 20 43 PM

It was better than this:

Screen Shot 2020-04-14 at 5 21 51 PM


And note that it only exhibits this mode on a 13 in with devtools on right. so it's in between phone (2 col) and desktop, where 3 cols shows all text

PR #176, opened 2020-04-15
Stop GH workflow duplicate runs

tstirrat:
There seems to be two runs on every PR which is killing my ocd.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -4,8 +4,10 @@
 name: Node.js CI
 
 on:
-  push: [master]
-  pull_request: [master]
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -4,61 +4,60 @@
 name: Node.js CI
 
 on:
-  - push
-  - pull_request
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
yuchenshi, line 10:
```suggestion
    branches: "*"
```

So we get CI when someone open a PR against a feature branch.

yuchenshi:
So I don't know if Renovate will still do its thing after merging this, let's find out! If it stops we can just whitelist branches that starts with the renovate prefix.

tstirrat:
> So I don't know if Renovate will still do its thing after merging this, let's find out! If it stops we can just whitelist branches that starts with the renovate prefix.

OK, lets monitor that

PR #178, opened 2020-04-15
RTDB JSON input type

tstirrat:
Adds a JSON field type:

Screen Shot 2020-04-21 at 5 50 35 PM
Screen Shot 2020-04-21 at 5 50 41 PM


src/components/Firestore/DocumentEditor/JsonEdtitor.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+  );
+};
+
+it('renders an editor for a document-ref', async () => {
+  const onChange = jest.fn();
+  const { getByLabelText, getByText } = render(
+    <TestForm>
+      <JsonEditor
+        name="foo"
+        value={fakeDocumentReference({
```
tjlav5, line 54: Should this just be a json blob?

src/components/Firestore/DocumentEditor/index.tsx
```diff
@@ -299,6 +303,14 @@ const FieldEditor: React.FC<{
         </div>
       </div>
     );
+  } else if (isJson) {
```
tjlav5, line 306: What do you think of separating json from the "primitive" fields? Then we don't need to have an isJson prop on these components

src/components/Firestore/DocumentEditor/JsonEdtitor.test.tsx
tstirrat: Yeah the tests are a copy paste job until I can get it selecting in the ui

src/components/Firestore/DocumentEditor/index.tsx
tjlav5: Right now the fieldEditor uses a hook to get a field from the store, and those fields can be map/array/primitive; with corresponding if-blocks in this component. What if we add a fourth field-type to the store (json) and add a fourth if-else block in this component that renders the JsonEditor?

tstirrat: That was my intuition but wasn't sure if it passed the sniff test of your design. Yes that sounds good to me!

tstirrat:
> Code LGTM, but can we make it a textarea instead?

`textarea` fails miserably with styling and sizing... :(


Screen Shot 2020-04-22 at 12 03 27 PM


I can certainly look into a follow up with resizable field rows, when I look at the other alignment issues.

PR #179, opened 2020-04-15
Enable missing doc navigation and creation

tstirrat:
Screen Shot 2020-04-15 at 4 10 05 PM

Screen Shot 2020-04-15 at 4 10 11 PM

Screen Shot 2020-04-15 at 4 10 17 PM

### DIFF
--- a/src/components/Firestore/DocumentPreview/api.tsx
+++ b/src/components/Firestore/DocumentPreview/api.tsx
@@ -49,6 +49,7 @@ export async function addFieldToMissingDocument(
 ) {
   const snapshot = await reference.get();
   if (snapshot.exists) {
+    // TODO: Better surface this to the user.
     throw new Error('.set() on an existing document would delete other fields');
   }
 
--- a/src/components/Firestore/DocumentPreview/api.tsx
+++ b/src/components/Firestore/DocumentPreview/api.tsx
@@ -42,6 +42,19 @@ export function updateField(
   documentRef.update(...payload);
 }
 
+export async function addFieldToMissingDocument(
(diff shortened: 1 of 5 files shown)


src/components/Firestore/DocumentPreview/index.tsx
```diff
@@ -67,14 +67,18 @@ const DocumentPreview: React.FC<Props> = ({
                 setIsAddingField(false);
               }}
               onSave={(key, value) => {
-                updateField(reference, data as FirestoreMap, [key], value);
+                if (!data) {
+                  reference.set({ [key]: value });
```
yuchenshi, line 71: Please moving this into `./api.ts`. Also worth considering adding a precondition to assert that the document actually doesn't exist to prevent accidentally overwriting data, but that can be in a follow up PR.

src/components/Firestore/DocumentPreview/api.tsx
```diff
@@ -42,6 +42,19 @@ export function updateField(
...
 }
 
+export async function addFieldToMissingDocument(
+  reference: firestore.DocumentReference,
+  key: string,
+  value: FirestoreAny
+) {
+  const snapshot = await reference.get();
+  if (snapshot.exists) {
+    throw new Error('.set() on an existing document would delete other fields');
```
yuchenshi, line 52:
```suggestion
    // TODO: Better surface this to the user.
    throw new Error('.set() on an existing document would delete other fields');
```

yuchenshi: We should definitely talk about the UX at some point, but until then... this TODO comment will do.

PR #185, opened 2020-04-20
Enable strict mode

tjlav5:


tjlav5:
> And it works? Sweet

There is one warning about RTDB using some legacy code, but it does work.

PR #186, opened 2020-04-21
Add pubsub and hosting cards to Home

tstirrat:
Screen Shot 2020-04-21 at 1 00 12 PM

Screen Shot 2020-04-21 at 3 38 20 PM

### DIFF
--- a/src/components/Home/index.tsx
+++ b/src/components/Home/index.tsx
@@ -88,6 +88,18 @@ const Overview: React.FC<{
           linkLabel="View Logs"
           testId="emulator-info-functions"
         />
+        <EmulatorCard
+          name="Hosting Emulator"
+          icon={<FunctionsIcon theme="secondary" />}
+          config={config.hosting}
+          testId="emulator-info-hosting"
+        />
+        <EmulatorCard
+          name="PubSub Emulator"
+          icon={<FunctionsIcon theme="secondary" />}
+          config={config.pubsub}
+          testId="emulator-info-pubsub"
+        />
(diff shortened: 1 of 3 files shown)


src/components/Home/index.tsx
```diff
@@ -88,6 +88,18 @@ const Overview: React.FC<{
           linkLabel="View Logs"
           testId="emulator-info-functions"
         />
+        <EmulatorCard
```
yuchenshi, line 91: I think we should link to the hosted website at host:port with a "View website" label.

tstirrat: 👍

tstirrat:
Screen Shot 2020-04-21 at 3 38 20 PM

src/components/Home/index.tsx
```diff
@@ -88,6 +88,22 @@ const Overview: React.FC<{
...
+          name="Hosting Emulator"
+          icon={<FunctionsIcon theme="secondary" />}
+          config={config.hosting}
+          testId="emulator-info-hosting"
+          linkToExternal={
+            config.hosting && `http://${config.hosting?.hostAndPort}/`
```
yuchenshi, line 97:
```suggestion
            config.hosting && `http://${config.hosting.hostAndPort}/`
```

I'm pretty sure TypeScript will narrow this down, no? Otherwise, use `config.hosting!.hostAndPort`.

tstirrat: Yeah, that works. thanks

PR #208, opened 2020-05-05
Tl/firestore query

tjlav5:
Adds the collection-filter functionality to the CollectionList... specifically filtering Documents that have certain field-attributes.

The main gist:
* Added a _store_ to Firestore, with actions/selectors
* The store maintains a CollectionFilter for a given CollectionPath for the life of the <Firestore /> component
* The panel is really just a fancy-form that tries to keep everything typesafe (not super easy with react-hook-form); and reads/writes to the store
* The Collection component conditionally adds _where_ and _orderBy_ clauses to the query based on values in the store 

The UI mostly stays true to the Console with the minor caveat that Console is essentially an accordion that only shows one panel at a time, this implementation shows any/all panels. If we get push back, we can add that functionality, but it will require some TLC for keyboard navigation/auto-focus.

Filtering with a "single"-value filter:
image

Filtering with a "multi"-value filter:
image

Collapsed panels with previews:
image

Badge when a filter is applied to the current collection-path:
image


src/components/Firestore/CollectionFilter.tsx
```diff
@@ -0,0 +1,455 @@
...
+
+  return (
+    <CollectionFilterTheme>
+      <FormContext {...(formMethods as any)}>
+        <form
+          onSubmit={formMethods.handleSubmit(onSubmit)}
+          className={className}
+        >
+          {/* Field entry */}
+          <FilterItem title="Filter by field" preview={cf.field} defaultOpen>
```
tstirrat, line 80: I love your layers of abstraction here. The fact that the button UI at the bottom is only slightly less lines compared to the content here for the main part of the form ❤️ 

Makes this easy to follow.

src/components/Firestore/CollectionFilter.tsx
```diff
@@ -0,0 +1,455 @@
...
+  if (isBoolean(value)) {
+    return 'boolean';
+  }
+  if (isNumber(value)) {
+    return 'number';
+  }
+  return 'string';
+}
+
+const ConditionEntry: React.FC<{
```
tstirrat, line 329: nit: Maybe extract the Condition* and Radio group components to new file(s).

This file is getting pretty massive.

src/components/Firestore/Collection.module.scss
```diff
@@ -0,0 +1,16 @@
+.query-panel {
```
tstirrat, line 1: Since we should move all to modules.. remove `.module.scss`, just `.scss`

Unless there's some specific reason it needs to be called that?

src/components/Firestore/CollectionFilter.tsx
```diff
@@ -0,0 +1,455 @@
...
+  name: string;
+  error?: string;
+}> = React.memo(({ name, error }) => {
+  const { setValue, watch } = useFormContext();
+  const value = watch(name);
+  const [fieldType, setFieldType] = useState(getConditionEntryType(value));
+
+  const numberRegex = /^-?([\d]*\.?[\d+]|Infinity|NaN)$/;
```
tstirrat, line 337: doesn't change, module level const

src/components/Firestore/CollectionFilter.tsx
tjlav5: Yeah I tend to agree with you, but do you think the right ™️  thing to do would be turn CollectionFilter into a folder with Condition* and Radio components? I'm not sold on adding those to the top-level Firestore/ directory.

src/components/Firestore/Collection.tsx
```diff
@@ -39,8 +49,35 @@ export interface Props {
...
+      collectionFilter.operator,
+      collectionFilter.values
+    );
+  }
+  if (collectionFilter && isSortableCollectionFilter(collectionFilter)) {
+    filteredCollection = filteredCollection.orderBy(
+      collectionFilter.field,
+      collectionFilter.sort
+    );
+  }
```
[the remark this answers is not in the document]
tjlav5: Good point, added to the bottom of this file.

src/components/Firestore/CollectionFilter.tsx
tjlav5: Thanks Tim!

src/components/Firestore/CollectionFilter.tsx
tstirrat: Yep, definitely +1 to grouping under a folder.

tjlav5: Cool, broke it up into smaller bits under a new CollectionFilter/ directory

PR #212, opened 2020-05-07
Add JSON import flow

tstirrat:
- Adds an import dialog
- Extracts out common rest api functionality from the two apis
- Creates a DatabaseApi, and RTDB ApiContext, useApi etc.
- Adds an overflow menu with JSON import dialog:

Screen Shot 2020-05-06 at 5 48 39 PM
Screen Shot 2020-05-06 at 4 58 59 PM
Screen Shot 2020-05-06 at 6 30 42 PM
Screen Shot 2020-05-06 at 5 48 55 PM


src/components/Database/DataViewer/ImportDialog.module.scss
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.container {
+  position: relative;
+
+  :global .Callout {
```
tjlav5, line 20: Does this need to be global? If it does, maybe it could live higher-up in the overrides?

tjlav5, line 20: Does this need to be global? If it does, maybe it could live higher-up in the overrides?

src/components/Database/Database.tsx
```diff
@@ -58,18 +74,35 @@ export const Database: React.FC<Props> = ({ config, namespace, path }) => {
...
           <InteractiveBreadCrumbBar
             base={urlBase}
             path={path || '/'}
             inputPrefix={getPrefix(ref)}
             onNavigate={handleNavigate}
-          />
+          >
+            <SimpleMenu
+              handle={<IconButton icon="more_vert" label="Open menu" />}
+              renderToPortal
```
tjlav5, line 87: Looking at the API, this seems like a prop for SimpleMenuSurface, not SimpleMenu. Could you confirm you need this?

src/components/common/Spinner.scss
```diff
@@ -1,5 +1,36 @@
...
+    left: 0;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    justify-content: center;
+
+    &.scrim {
+      background-color: $COLOR_WHITE_ALPHA_87;
+    }
+  }
```
tjlav5, line 35: 💅

src/components/Database/DataViewer/ImportDialog.module.scss
tstirrat: `:global` means that it will not prefix that immediate class, so it's still local to this scoped component... is weird. if you nest styles in scss + modules it prefixes them all!

Without global: `.container_xy .container_xy_Callout`, which seems totally useless!

With global: `.container_xy .Callout`...

tstirrat: https://github.com/webpack-contrib/css-loader/issues/346

I preferred this over adding a specific class to `<Callout>` just to add margin... 

It's really the container who cares about the margin.

I don't love this..

In fact this is making me lean towards [emotion styled-components](https://emotion.sh/docs/nested), which support style encapsulation AND nice nesting ergonomics:


```ts
const Container = styled.div`
  color: red;

  // nesting other raw styles works fine
  .Callout {
    margin-bottom: 24px;
  }
  
  // nesting other styled components:
  ${SomeOtherComponent} {
    margin: 8px;
  }
`;
```

src/components/Database/Database.tsx
tstirrat: It's needed otherwise this wraps:

Screen Shot 2020-05-07 at 10 23 33 AM

src/components/Database/DataViewer/ImportDialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+import styles from './ImportDialog.module.scss';
+
+export interface Props {
+  api: DatabaseApi;
+  reference: firebase.database.Reference;
+  onComplete: (reference?: firebase.database.Reference, file?: File) => void;
+}
+
+export const ImportDialog: React.FC<Props> = ({ reference, onComplete }) => {
+  const api = useApi();
```
yuchenshi, line 42: Please remove this and the `useApi()` hook + Context. This component already takes `api` in the props, and there is no reason to introduce a Context. Also general comment: I'd strongly prefer passing a prop if a prop will do, since it is straightforward to reason about, statically type-checked, and does not throw runtime errors.

src/components/Database/DataViewer/ImportDialog.module.scss
yuchenshi: Tim is right about this `:global` usage. However, this line can be made much less ambiguous:
```suggestion
  :global(.Callout) {
```

This clarifies that the class name `Callout` is global while the whole rule is not.

src/components/Database/DataViewer/ImportDialog.tsx
tstirrat: I thought it might be needed "lower down" later... but I can remove for now.

yuchenshi: Yeah, my advice would be keep "drilling" the property until at some point we decide it's too annoying, and migrate to Context. We followed the route for Firestore viewer and now it's in Context-land, but I wouldn't recommend introducing a Context "just in case" before drilling drags us down to much. It's easy to migrate drilling to Context but the opposite is much harder due to the dynamic nature.

tstirrat:
I added error handling:

Screen Shot 2020-05-07 at 12 05 58 PM

PR #217, opened 2020-05-07
Add validation for collection-filter

tjlav5:
image

### DIFF
--- a/src/components/Firestore/CollectionFilter/ConditionEntry.tsx
+++ b/src/components/Firestore/CollectionFilter/ConditionEntry.tsx
@@ -31,71 +31,77 @@ function getConditionEntryType(value: any) {
   return 'string';
 }
 
-export const ConditionEntry: React.FC<{
+interface ConditionEntryProps {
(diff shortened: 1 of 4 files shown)


src/components/Firestore/CollectionFilter/ConditionEntry.tsx
```diff
@@ -33,7 +33,7 @@ function getConditionEntryType(value: any) {
 
 export const ConditionEntry: React.FC<{
   name: string;
-  error?: string;
+  error: string | undefined;
```
tstirrat, line 36: is this logically equivalent change needed? if so, because it's unusual, add a comment. My instinct would be to change it back if I saw it in passing.

yuchenshi: If you want to require the `error` field to always be specified, but may be `undefined`, please add a comment. (BTW, this is totally acceptable and reasonable since it reduces human mistakes of forgetting to pass `error` through.)

PR #221, opened 2020-05-08
Update Home/icons. Hide action-divider when there are no actions.

tjlav5:
image

Keeps the actions, just hides the divider; then all cards still have the same height - as opposed to removing the actions-div which results in a shorter-card.


src/components/common/icons.tsx
```diff
@@ -66,6 +66,27 @@ export const FunctionsIcon = svgIcon(
...
+  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 192 192">
+    <path
+      d="M96 16c-44.16 0-80 35.84-80 80s35.84 80 80 80 80-35.84 80-80-35.84-80-80-80zm-8 143.44c-31.6-3.92-56-30.8-56-63.44 0-4.96.64-9.68 1.68-14.32L72 120v8c0 8.8 7.2 16 16 16zm55.2-20.32c-2.08-6.48-8-11.12-15.2-11.12h-8v-24c0-4.4-3.6-8-8-8H64V80h16c4.4 0 8-3.6 8-8V56h16c8.8 0 16-7.2 16-16v-3.28c23.44 9.52 40 32.48 40 59.28 0 16.64-6.4 31.76-16.8 43.12z"
+      fillRule="evenodd"
+      fill="currentColor"
+    />
+  </svg>
+);
+
+export const PubSubIcon = svgIcon(
```
yuchenshi, line 79: no `fill="currentColor" on any of the path / circles?

PR #234, opened 2020-05-13
Change SASS_PATH to be more universal

tjlav5:


tjlav5:
Ah you're right, that should have been a semicolon...

tjlav5:
Doesn't work with a semicolon

PR #237, opened 2020-05-13
fix(Realtime Database): Apply filters to table views

davideast:
Filters did not previously work on Table Views due to reading from reference instead of query.

image

image

### DIFF
--- a/src/components/Database/DataViewer/common/fetch.ts
+++ b/src/components/Database/DataViewer/common/fetch.ts
@@ -90,7 +90,7 @@ function realtimeToViewModel(
   queryParams: QueryParams
 ) {
   const query = applyQuery(ref, queryParams);
-  return once(ref).pipe(
+  return once(query).pipe(
     switchMap(snap => {
       return snap.hasChildren() ? toObservable(query) : toObservable(ref);
     }),
@@ -100,6 +100,7 @@ function realtimeToViewModel(
         children: snap.hasChildren() ? Object.keys(snap.val()) : NO_CHILDREN,
         isRealtime: true,
         isLoading: false,
+        query,
       })
     )
   );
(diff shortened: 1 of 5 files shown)


src/components/Database/DataViewer/common/fetch.ts
```diff
@@ -90,7 +90,7 @@ function realtimeToViewModel(
   queryParams: QueryParams
 ) {
   const query = applyQuery(ref, queryParams);
-  return once(ref).pipe(
+  return once(query).pipe(
```
yuchenshi, line 93: Code LGTM but I don't get the reason behind this section, especially the switchMap below, and I'd like to hear some thoughts from @tstirrat .

tstirrat: Ok will talk offline about it.

tstirrat: I actually think you should remove this line change. This is an initial detection to work out if a filter can be applied at the node.. so it's better to `once(ref)`.. if the once(query) returned no nodes, it would then give strange results.

tstirrat: If you filtered to zero children, then it would yield `snap.hasChildren() === false`, and then it would return `toObservable(ref)` with the ref (not query).. so filters that yield no results will **inactivate themselves**.

src/components/Database/DataViewer/common/fetch.ts
```diff
@@ -89,17 +89,25 @@ function realtimeToViewModel(
   ref: firebase.database.Reference,
   queryParams: QueryParams
 ) {
-  const query = applyQuery(ref, queryParams);
+  let query: firebase.database.Query;
```
tstirrat, line 92: I reverted the `once(query)` because it would have yielded strange results..

The check is really: 

does the **NODE** have children?
- if so, apply query
- else, it's a leaf so no queries.

If we left it as `once(query)` it would have done bad things:

Does the query have children?
- yes: all good
- no: return the full unfiltered node -- this would mean that queries that yield no results magically return everything.

PR #239, opened 2020-05-14
Allow clicking ids (keys) in table view

tstirrat:


yuchenshi:
I remember last time we've talked about this -- basically the only concern is that dependency on router deep down may make it harder to extract RTDB viewer as a standalone reusable component. If you've changed your mind since then, then I won't block either.

tstirrat:
> I remember last time we've talked about this -- basically the only concern is that dependency on router deep down may make it harder to extract RTDB viewer as a standalone reusable component. If you've changed your mind since then, then I won't block either.

Ah... I guess YAGNI for now and we could revisit that and plumb down if that became a distinct project

yuchenshi:
Okay then. Tests please :)

tstirrat:
> Okay then. Tests please :)

Done. Added tests, and I made it a little more `<Route>` compatible

src/components/Database/DataViewer/NodeLink.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+  it('appends ref path to root routes', () => {
+    const { getByText } = render(
+      <MemoryRouter initialEntries={['/x/y/z']}>
+        <Route path="/">
+          <NodeLink dbRef={REF} />
+        </Route>
+      </MemoryRouter>
+    );
+
+    expect(getByText('my_key').href).toEqual(`http://a/b/c/my_key`);
```
yuchenshi, line 85: Huh what? No hostname?

Also, remember that you can test the value of the `href` HTML attribute by `.getAttribute("href")` which gives you the attribute as-is, so you don't have to deal with absolute paths.

src/components/Database/DataViewer/NodeLink.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const { getByText } = render(
+      <MemoryRouter initialEntries={['/database/my-instance/data/x/y']}>
+        <Route path="/database/:instance/data/:path*">
+          <NodeLink dbRef={REF} />
+        </Route>
+      </MemoryRouter>
+    );
+
+    expect(getByText('my_key').getAttribute('href')).toEqual(
+      `/database/my-instance/data/a/b/c/my_key`
```
yuchenshi, line 58: nit: please make these strings normal strings instead of template strings. More of those below.

tstirrat: Done.

PR #240, opened 2020-05-14
View only functions logs with /functions.

yuchenshi:
This makes `http://localhost:4000/functions` show function logs, so that https://github.com/firebase/firebase-tools/pull/2219 can link to it.

### DIFF
--- a/src/components/LogViewer/index.tsx
+++ b/src/components/LogViewer/index.tsx
@@ -21,7 +21,7 @@ import { Elevation } from '@rmwc/elevation';
 import { GridCell } from '@rmwc/grid';
 import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
-import { useLocation, useRouteMatch } from 'react-router-dom';
+import { useLocation } from 'react-router-dom';
 
 import { AppState } from '../../store';
 import { LoggingConfig } from '../../store/config';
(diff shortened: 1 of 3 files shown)


src/components/LogViewer/index.tsx
```diff
@@ -50,12 +51,17 @@ export type Props = PropsFromState & PropsFromDispatch;
 
 const compiledGetters = new CompiledGetterCache();
 
+function useQuery() {
+  return new URLSearchParams(useLocation().search);
```
tstirrat, line 55: `new UrlSearchParams` will return a new reference each time, so I think this could cause weirdness even though useLocation() should have hook state mechanics...

I suspect you want `useEffect(() => ..., [location])` or useMemo

yuchenshi: I basically got this from the official react router docs https://reacttraining.com/react-router/web/example/query-parameters. I've added `useMemo` as an optimization, but I can assure you that there is no weirdness either way, unless the new search params object is passed to a child component which checks props by reference.

Hooks are just functions. If we're fine with `parseSomething(string)` in render then it's not that different for `new UrlSearchParams(string)`. `useMemo` can then be applied as [optimization but not semantic guarantee](https://reactjs.org/docs/hooks-reference.html#usememo). And I don't think `useEffect` is right here -- we just want a simple value to pass to another hook and there is no side effect.

tstirrat: Yes, that's what I was worried about: `query.get("name")` is fine but passing down `query` could cause unnecessary re-renders.

Thanks for updating it!

abeisgoat:
thansk guys!

Issue #243, opened 2020-05-14
FR: Use absolute paths for cloning in RTDB


davideast:
Cloning could allow a user to specify an absolute path for easier database restructuring. 

![image](https://user-images.githubusercontent.com/4570265/81979027-95b7f100-95e9-11ea-8a83-c62a4fce1a27.png)

PR #247, opened 2020-05-14
feat(Realtime Database): Clone with absolute paths

davideast:
Addresses #243.

Change CloneDialog to accept absolute URLs rather than only allow cloning within the parent node.

The CloneDialog now shows an absolute path in the `newKey` input to indicate you now are cloning with an absolute path.

image

### DIFF
--- a/src/components/Database/DataViewer/CloneDialog.tsx
+++ b/src/components/Database/DataViewer/CloneDialog.tsx
@@ -39,8 +39,26 @@ export interface Props {
   onComplete: (string?: string) => void;
 }
 
+/**
+ * Strips protocol and domain name (usually http://localhost:9000/)
+ * from the ref's path
+ * @example "localhost:9000/todos/one" -> "/todos/one"
+ * @param ref
+ */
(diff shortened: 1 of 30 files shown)


src/components/Database/DataViewer/NodeActions.tsx
```diff
@@ -34,6 +34,7 @@ import { RenameDialog } from './RenameDialog';
 
 export interface Props {
   realtimeRef: firebase.database.Reference;
+
```
tstirrat, line 37: remove diff

src/components/Database/DataViewer/CloneDialog.tsx
```diff
@@ -83,7 +102,7 @@ export const CloneDialog = React.memo<Props>(function CloneDialog$({
     Object.entries(form).forEach(([key, value]) => {
       data[key] = JSON.parse(value);
     });
-    realtimeRef.parent!.child(newKey).set(data);
+    realtimeRef.root.child(newKey).set(data);
```
tstirrat, line 105: Now that its a path and not a key, update the UI to say something like `Destination (path)`

PR #256, opened 2020-05-18
RTDB: Show whitespace in keys

tstirrat:
Screen Shot 2020-05-18 at 1 53 32 PM

http://b/152781501

### DIFF
--- a/src/components/Database/DataViewer/NodeLink.tsx
+++ b/src/components/Database/DataViewer/NodeLink.tsx
@@ -38,7 +38,7 @@ export const NodeLink = React.memo<Props>(function NodeLink$({ dbRef }) {
       tag={Link}
       to={href}
     >
-      <span className="key">{key}</span>
+      {key}
     </Typography>
   );
 });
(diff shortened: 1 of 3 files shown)


src/components/Database/DataViewer/NodeLink.tsx
```diff
@@ -27,7 +27,7 @@ export interface Props {
 }
 
 export const NodeLink = React.memo<Props>(function NodeLink$({ dbRef }) {
-  const key = dbRef.parent === null ? getDbRootUrl(dbRef) : dbRef.key;
+  const key = dbRef.key || getDbRootUrl(dbRef);
```
yuchenshi, line 30: Does RTDB allow a key that's the empty string?

tstirrat: Good Q, I don't think so. but let's check

tstirrat: Nope, phew!

```
database.ref().child('///a//').key
// "a"

database.ref().child('///a//').toString()
"http://localhost:9000/a"
```

PR #258, opened 2020-05-19
Firestore: Encode document id on URI

edupsousa:
Small fix to show firestore documents that contains # and other special characters on the id.

### DIFF
--- a/src/components/Firestore/Collection.tsx
+++ b/src/components/Firestore/Collection.tsx
@@ -135,7 +135,7 @@ export const Collection: React.FC<Props> = ({ collection }) => {
               key={doc.ref.id}
               className="Firestore-List-Item"
               tag={NavLink}
-              to={`${url}/${doc.ref.id}`}
+              to={`${url}/${encodeURIComponent(doc.ref.id)}`}
               activeClassName="mdc-list-item--activated"
             >
               {doc.ref.id}
@@ -146,7 +146,8 @@ export const Collection: React.FC<Props> = ({ collection }) => {
       <Route
         path={`${url}/:id`}
         render={({ match }: any) => {
-          const docRef = collection.doc(match.params.id);
+          const docId = decodeURIComponent(match.params.id);
+          const docRef = collection.doc(docId);
           return <Document reference={docRef} />;
         }}
       ></Route>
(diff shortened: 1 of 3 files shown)


edupsousa:
@googlebot I signed it!

src/components/Firestore/Collection.tsx
```diff
@@ -146,7 +146,8 @@ export const Collection: React.FC<Props> = ({ collection }) => {
       <Route
         path={`${url}/:id`}
         render={({ match }: any) => {
-          const docRef = collection.doc(match.params.id);
+          const docId = decodeURIComponent(match.params.id);
```
yuchenshi, line 149: I think react router should already decode this for us, no? Please double check to make sure we're not double decoding.

src/components/Firestore/Collection.tsx
```diff
@@ -135,7 +135,7 @@ export const Collection: React.FC<Props> = ({ collection }) => {
               key={doc.ref.id}
               className="Firestore-List-Item"
               tag={NavLink}
-              to={`${url}/${doc.ref.id}`}
+              to={`${url}/${encodeURIComponent(doc.ref.id)}`}
```
tstirrat, line 138: Can you write a test for this? Thanks

src/components/Firestore/api.tsx
```diff
@@ -58,7 +58,10 @@ export default class DatabaseApi extends RestApi implements FirestoreApi {
   private async getSubCollections(
     docRef: firestore.DocumentReference
   ): Promise<firestore.CollectionReference[]> {
-    const encodedPath = docRef.path; // TODO: Encode each segment.
+    const encodedPath = docRef.path
```
tstirrat, line 61: A test for this would be helpful too

src/components/Firestore/Collection.tsx
minho-comcom-ai: @edupsousa I added a document with percent-encoded '/'(slash) included at Id. (e.g. 'hello%2fworld')
And it finally works without `decodeURIComponent()` in this line.
(@minhoryang)

yuchenshi:
This is now out of date due to the `api.tsx` changes. If anyone still has bandwidth on this, feel free to update the PR or send us a new one. We just need to move the path encoding logic to the new files and add some tests.

yuchenshi:
This is merged as https://github.com/firebase/firebase-tools-ui/pull/439. Thanks for the contribution!

PR #296, opened 2020-06-11
Group @rmwc packages in renovate config

tstirrat:


yuchenshi:
I'm not sure if we want this.  I remember that we still cannot upgrade `@rmwc/select` but other components upgrade just fine. I'm down if you want to rate limit it a bit to reduce spam but I think it's better to keep the granularity.

tstirrat:
Ah yes I see it could cause them to break and not be mergeable. ala textfield right now.

PR #303, opened 2020-06-13
add ctrl+k to clear logs

dutzi:
This PR adds the options to hit Ctrl+K to clear the function's logs history.

I hope this gets accepted, let me know if you want a UI for it, I was going for it initially but didn't want to make hasty decisions.


abeisgoat:
I'm not opposed to adding this, per se, but I'm more inclined to add a clear button (discoverable, obvious) over an undocumented keyboard shortcut. I'll bring this up in our weekly sync and see how folks feel.

PR #353, opened 2020-09-10
Firestore: Add a copy ID button

davideast:
Adds a "Copy ID" `<IconButton />` to a `<PanelHeader />`.

copyid

## Questions
1. Should the `<CopyButton />` but on the left or right side of the `<PanelHeader />`? I recommend the left side to keep the menus and filters in their same places. The right most is usually the primary action as well.


yuchenshi:
2 => no copy button for root please
3 => Clipboard API is fine, as long as we hide the button in browsers that don't have Clipboard API (feature detection and progressive enhancement)
4 => the button is definitely more intuitive

src/components/common/CopyButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  React.useEffect(() => reset.current, [text]);
+
+  return {
+    copied,
+    copy: React.useCallback(
+      () =>
+        navigator.clipboard
+          .writeText(text)
+          .then(() => setCopied(true))
+          .catch(() => setCopied(copied => copied)),
```
NothingEverHappens, line 22: Seems like the type of copied is boolean, but you're setting a callback here, is this intended?

src/components/common/CopyButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+import React from 'react';
+
+interface CopyButtonProps {
+  text: string;
+  label?: string;
+}
+
+// TODO(davideast): Depend on pkg https://github.com/jaredLunde/react-hook/tree/master/packages/copy#readme
+function useCopy(text: string) {
+  const [copied, setCopied] = React.useState(false);
```
NothingEverHappens, line 11: Looks like we import useState/useRef directly everywhere else in the app, e.g. `useState(false);`

src/components/common/CopyButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const reset = React.useRef(() => setCopied(false));
+  React.useEffect(() => reset.current, [text]);
```
NothingEverHappens, line 13: doesn't seem like reset is used anywhere? Safe to drop?

NothingEverHappens (approved):
I'm not super familiar with the firestore code

src/components/common/CopyButton.tsx
davideast: This was a port of [`useCopy()`](https://github.com/jaredLunde/react-hook/tree/master/packages/copy). I didn't think we would use our own version but I didn't want to bring in a dependency before getting approval. So ignore this hook unless you want us to use our own.

davideast:
Updated the main description with the new GIF and remaining question.

yuchenshi (approved):

PR #372, opened 2020-10-09
Feat auth server error handling + final polish for beta

NothingEverHappens:


src/components/Auth/UserFormDialog/UserForm.test.tsx
```diff
@@ -30,93 +37,117 @@ describe('UserForm', () => {
...
-    expect((getByLabelText(/Display name/) as HTMLInputElement).value).toBe(
-      displayName
-    );
-    expect((queryByLabelText('Email') as HTMLInputElement).value).toBe(email);
-    expect((queryByLabelText('Password') as HTMLInputElement).value).toBe(
-      password
-    );
+
+    expect(createUser).toHaveBeenCalled();
+    expect(clearAuthUserDialogData).toHaveBeenCalled();
```
yuchenshi, line 139: Please add an assertion on the dialog still open

src/components/Auth/UserFormDialog/UserForm.tsx
```diff
@@ -1,85 +1,145 @@
...
-            type="button"
-          >
-            Save and create another
-          </Button>
-        )}
+          <ImageUrlInput {...form} />
+          <CustomAttributes {...form} />
+          <SignInMethod {...form} user={user} />
+          {hasError(authUserDialogData?.result) && (
+            <Callout type="warning">Unknown server error</Callout>
```
yuchenshi, line 93: Have we considered surfacing the error? It will make debugging easier (or at least make future GitHub issues less cryptic)

src/store/auth/actions.tsx
```diff
@@ -76,3 +77,19 @@ export const getAllowDuplicateEmailsRequest = createAction(
 export const getAllowDuplicateEmailsSuccess = createAction(
   '@auth/GET_ALLOW_DUPLICATE_EMAILS_SUCCESS'
 )<boolean>();
+
+export const setAuthUserDialogData = createAction(
```
yuchenshi, line 81: Can we have a more high-level action like `openAuthUserDialog` instead?

src/store/auth/selectors.ts
```diff
@@ -3,14 +3,31 @@ import { createSelector } from 'reselect';
...
 });
 
+export const getAuthUserDialog = createSelector(getAuth, (state: AuthState) => {
+  return state.authUserDialogData;
+});
+
+export const getCurrentEditedUser = createSelector(
+  getAuthUserDialog,
+  authUserDialogData => {
+    return authUserDialogData && hasData(authUserDialogData!.result)
```
yuchenshi, line 21: nit: Do we need `!` here? I think TypeScript should already narrow it down

src/store/auth/selectors.ts
```diff
@@ -3,14 +3,31 @@ import { createSelector } from 'reselect';
...
+      ? authUserDialogData!.result.data
```
yuchenshi, line 22: Same

yuchenshi (approved):

src/components/Auth/UserFormDialog/UserForm.test.tsx
yuchenshi: Also: there are reports of the dialog got closed in the bugbash. Please investigate

NothingEverHappens: I moved the dialog moving logic out of the form and fixed the bug where the dialog closed.

src/components/Auth/UserFormDialog/UserForm.tsx
NothingEverHappens: Going to show the error message coming from the server

src/components/common/rest_api.ts
```diff
@@ -52,7 +52,7 @@ export abstract class RestApi {
     const json = await res.json();
 
     if (res.status >= 400 && res.status < 600) {
-      throw new ResponseError(await res.text(), res, json);
+      throw new ResponseError(await json.error.message, res, json);
```
yuchenshi, line 55:
```suggestion
      throw new ResponseError(json.error.message, res, json);
```

src/components/common/rest_api.ts
```diff
@@ -71,7 +71,7 @@ export abstract class RestApi {
...
```
yuchenshi, line 74:
```suggestion
      throw new ResponseError(json.error.message, res, json);
```

src/store/auth/selectors.ts
```diff
@@ -71,7 +71,8 @@ export const getFilteredUsers = createSelector(
     return users.filter(u => {
       return [u.localId, u.displayName, u.email || '', u.phoneNumber || '']
         .join('\n')
-        .includes(filter);
+        .toLocaleLowerCase()
```
yuchenshi, line 74: Please use `toLocaleUpperCase` instead (and same below) due to l10n concerns: https://stackoverflow.com/a/14128850

(Side note: It's still not perfect, but better than lowering the case. The best way would be using a locale-aware method, but I'm not _aware_ of such a method for `.includes`.)

yuchenshi (approved):

src/store/auth/selectors.ts
NothingEverHappens: Looks like this only applies if we want to convert it back
I also have 0 preference here, so switched to upperCase

PR #425, opened 2020-11-20
Always include projectId in Auth APIs.

yuchenshi:
It may still talk to the wrong projectId under certain broken setups, but at least it's consistently wrong now.


davideast (approved):
LGTM.

Thanks for making this obvious when broken.

PR #435, opened 2020-11-25
Enable concurrency-mode

tjlav5:
Switch to react/experimental and bootstrap with unstable_createRoot


src/index.tsx
```diff
@@ -48,7 +48,7 @@ const RouterWithInit = () => {
   );
 };
 
-ReactDOM.render(
+ReactDOM.unstable_createRoot(document.getElementById('root')!).render(
```
yuchenshi, line 51: Please only enable blocking mode for now.

yuchenshi (commented):
LGTM but please switch to blocking mode. Also, let's set up a branch for these changes -- we have a few bug fixes to the stable branch that I want to get out before going full on experimental

jhuleatt:
Heads up that ReactFire v3 will support apps that don't use concurrent mode, which means you can stick to a stable build of React. More detail on the API changes that enable this here: https://github.com/FirebaseExtended/reactfire/pull/255#issue-417561976

yuchenshi:
I guess we're already using `Suspense` for data fetching so switching to at least blocking mode is inevitable. In retrospect, in an alternative world where we had ReactFire v3 a few months ago, we could have decided not to hop on the Suspense train (pun intended). Closing in favor of #440.

PR #439, opened 2020-12-01
Firestore: Encode special characters inside document ids on URI

AntonioRdzNav:
This is an update on the [pull request #258](https://github.com/firebase/firebase-tools-ui/pull/258), this version is working with the latest of master.

However, I found out that the PR 258 had a bug when encoding the URI if the document id had any '%' character on it. The bug is a known issue of the **history** library that can be found in [here.](https://github.com/ReactTraining/history/issues/505)

The '%' bug is currently solved by a simply workaround (removing '%' characters before encoding the URI), so a better solution would be appreciated.

All other special characters are correctly encoded by encodeURIComponent, see [this guide](https://thisthat.dev/encode-uri-vs-encode-uri-component/) as reference to which special characters are encoded by encodeURIComponent and which ones are not.

Unit Tests are pending to track current behavior, all other Unit Tests are passed with no problems.

Finally, this PR only focuses on encoding document ids, but collection ids are not yet being encoded.


AntonioRdzNav:
@googlebot I consent.

src/components/Firestore/FirestoreEmulatedApiProvider.tsx
```diff
@@ -119,7 +119,10 @@ export function useSubCollections(
   docRef: firebase.firestore.DocumentReference
 ) {
   const { baseUrl } = useFirestoreRestApi();
-  const encodedPath = docRef.path; // TODO: Encode each segment
+  const encodedPath = docRef.path
+    .split('/')
+    .map(uri => encodeURIComponent(uri.replace(/%/g, '')))
```
yuchenshi, line 124: There is no need to remove % here -- the Firestore Emulator, unlike React router, handles it just fine.

yuchenshi (changes requested):
The workaround silently changes the document path so developers may see a different document or a confusing "document not exist" message. Personally, I'd rather just let it error and file a bug to track it.

AntonioRdzNav:
With the latest commits, the workaround of the % bug was removed, so filing a bug should be the next step.

The additional changes were:
- Encoded and decoded the collection ids as well (instead of just the document ids).
- Implemented Unit Tests to track every encode/decode behaviors.

src/components/Firestore/Collection.test.tsx
```diff
@@ -227,9 +227,13 @@ it('shows the selected sub-document', async () => {
 });
 
 describe('withCollectionState', () => {
-  let performAddDocument: () => Promise<void>;
+  let performAddDocument: (hasSpecialId?: boolean) => Promise<void>;
```
yuchenshi, line 230: nit: Boolean arguments are hard to read in general -- consider changing it to `performAddDocument(id = 'new-document-id')` and callers can pass in any id (including special characters) if they want to.

src/components/Firestore/Collection.test.tsx
```diff
@@ -284,4 +288,32 @@ describe('withCollectionState', () => {
...
+        );
+      },
+      {
+        path: '/my-stuff/my-doc/sub-coll',
+      }
+    );
+
+    await waitForElement(() => getByTestId('withCollectionState'));
+
+    await act(() => performAddDocument(true));
```
yuchenshi, line 315: See above

yuchenshi:
This is getting better -- don't worry about the CLA check, I can handle that later.

If you can file an issue on the % issue and add a test for it, that'll be great. Make sure to link to the issue on the test. The test should show the ideal behavior with % and it should fail for now -- and we can mark it as `it.skip` so it doesn't break CI.

AntonioRdzNav:
> This is getting better -- don't worry about the CLA check, I can handle that later.
> 
> If you can file an issue on the % issue and add a test for it, that'll be great. Make sure to link to the issue on the test. The test should show the ideal behavior with % and it should fail for now -- and we can mark it as `it.skip` so it doesn't break CI.

I implemented several Unit Tests that test encode and decode behaviors in different files, would it be better if I add an ideal behavior test **per file**? or would just one ideal behavior test in **one of the files** be enough?

yuchenshi (approved):

PR #451, opened 2020-12-07
Add lexicographical order for fields in document

### LINKED ISSUE #438: Firestore document fields, subcollections and map keys are not in lexicographical order
_I haven't found any issue template, please tell me if I need to edit this._

### Problem
I've noticed that in the local firestore UI the fields of a document, its subcollections and the fields of a map inside the document are not lexicographically ordered.
actual

### To reproduce
1. Startup the firestore emulator.
1. Create a new collection `temp`
1. Create a new doc `temp`
1. Add fields and fields in non-lexicographical order, e.g `Z: "Z"`, `F: "F"`, `A: "A"`,`G: "G"`.
1. They are ordered by the added time, not lexicographically

### Expected behavior
In the Firestore web UI the lexicographical order is followed.
expected

### Motivation
I believe it should be that way to give consistency between the two environments, to help also the developer when confronting the 2 DBs.

### Fix
It should not be difficult to fix by adding a sort method when mapping the keys to the great components. I would be very happy to do it.

Thank you

GabriFila:
This PR adds the lexicographical ordering for the fields in a document on the UI. I didn't touch the files related to collections and subcollections since I noticed they were already ordered.

I also adapted the tests to the new ordering. Do I need to add also a test that clearly checks for the ordering? If so, do you have any suggestions on how to check that an element comes after/before another one? Thank you!

The following images show the before and after, where in the after the order is always respected

#### before
befFields
#### after
aftFields
#### before
bef nested
#### after
aftNested
#### before
befSum
#### after
aftSum

Fixes #438

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,7 @@
 .eslintcache
 settings.json
 
-*-debug.log*
+*debug.log*
 yarn-error.log*
 
 .idea
--- a/src/components/Firestore/DocumentPreview/FieldPreview.tsx
+++ b/src/components/Firestore/DocumentPreview/FieldPreview.tsx
@@ -27,6 +27,7 @@ import React, { useState } from 'react';
 import { supportsEditing } from '../DocumentEditor';
 import { FirestoreArray } from '../models';
 import {
+  compareFirestoreKeys,
   getFieldType,
   getParentPath,
   isArray,
@@ -54,17 +55,19 @@ const FieldPreview: React.FC<{
   let childFields = null;
   if (isMap(state)) {
     // Inline editor for new field will default to key: ''
-    childFields = Object.keys(state).map(childLeaf => {
-      const childPath = [...path, childLeaf];
-      return (
-        <FieldPreview
-          key={childLeaf}
-          path={childPath}
-          documentRef={documentRef}
-          maxSummaryLen={maxSummaryLen}
-        />
-      );
-    });
+    childFields = Object.keys(state)
+      .sort(compareFirestoreKeys)
+      .map(childLeaf => {
+        const childPath = [...path, childLeaf];
+        return (
+          <FieldPreview
+            key={childLeaf}
+            path={childPath}
+            documentRef={documentRef}
+            maxSummaryLen={maxSummaryLen}
+          />
+        );
+      });
   } else if (isArray(state)) {
     // Inline editor for new field will default to key: '{index}'
     childFields = state.map((value, index) => {
--- a/src/components/Firestore/DocumentPreview/index.test.tsx
+++ b/src/components/Firestore/DocumentPreview/index.test.tsx
@@ -279,8 +279,8 @@ describe('loaded map', () => {
       documentReference.update = jest.fn();
       await documentReference.set({
         foo: {
-          first_name: 'harry',
           last_name: 'potter',
+          first_name: 'harry',
         },
       });
       return (
@@ -291,10 +291,10 @@ describe('loaded map', () => {
     await waitForElement(() => result.getAllByText(/harry/).length > 0);
   });
 
-  it('renders a field', () => {
+  it('renders a field, in the lexicographical order', () => {
     const { getByText } = result;
 
-    expect(getByText('{last_name: "potter", first_name: "harry"}')).not.toBe(
+    expect(getByText('{first_name: "harry", last_name: "potter"}')).not.toBe(
       null
     );
     expect(getByText('(map)')).not.toBe(null);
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -23,5 +23,11 @@ settings.json
...
 *-debug.log*
 yarn-error.log*
+debug.log
 
 .idea
+
+# firestore emulator
+
+# vscode local history plugin
+.history
```
yuchenshi, line 33: Please kindly revert changes here. Also, a word in general: please [use a global `.gitignore`](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) for your plugins. I apologize for the inconvenience, but including each and every IDE plugin in a repo won't scale well.

src/components/Firestore/DocumentPreview/FieldPreview.tsx
```diff
@@ -54,17 +54,19 @@ const FieldPreview: React.FC<{
...
-        <FieldPreview
-          key={childLeaf}
-          path={childPath}
-          documentRef={documentRef}
-          maxSummaryLen={maxSummaryLen}
-        />
-      );
-    });
+    childFields = Object.keys(state)
+      .sort((fieldA, fieldB) => fieldA.localeCompare(fieldB))
```
yuchenshi, line 58: Would you mind extracting a function like `compareFirestoreKeys` for all uses below? That function can live in `util.ts` or somewhere close. We may need to change the comparison or apply polyfills globally.

(Side note: Firestore uses a different sorting method than `localeCompare`, but I'd still suggest we merge this for now since it is close enough for most browsers in English locales and we can improve it later.)

src/components/Firestore/DocumentPreview/index.test.tsx
```diff
@@ -306,7 +306,7 @@ describe('loaded map', () => {
     expect(getByText('first_name')).not.toBe(null);
     expect(getByText('last_name')).not.toBe(null);
 
-    getByText('{last_name: "potter", first_name: "harry"}').click();
+    getByText('{first_name: "harry", last_name: "potter"}').click();
```
yuchenshi, line 309: note: This change is good enough for test coverage. There is no need to add another test case for now.

yuchenshi (commented):
Mostly LGTM. See comments inline.

.gitignore
GabriFila: Thank you for the amazing suggestion, did not know about that

src/components/Firestore/DocumentPreview/FieldPreview.tsx
GabriFila: Ok! Do you want to use that function just in that place, or use it also for maps inside firestore documents?

src/components/Firestore/DocumentPreview/index.test.tsx
GabriFila: Great! Thanks

src/components/Firestore/DocumentPreview/FieldPreview.tsx
yuchenshi: Please use the function for all new sorting you've introduced, thanks! If you notice any other places where we're sorting keys, feel free to refactor those too.

As far as string comparison goes, Firestore production uses UTF-8 encoded byte order ([docs](https://firebase.google.com/docs/firestore/manage-data/data-types)). Unfortunately, it is very hard to get that right in client-side JavaScript (even if you use [hacks](http://ecmanaut.blogspot.com/2006/07/encoding-decoding-utf8-in-javascript.html)), so please don't stress yourself to get that in this PR. A easily swappable function plus useful comments that explains the situation is good enough.

GabriFila: Ok, I will not waste time on that since it seems pretty tricky. THanks for the info!

GabriFila:
@yuchenshi Can I add the `Fix` word where I cite the issue in the PR description to automatically close it?

src/components/Firestore/utils.ts
```diff
@@ -213,3 +215,7 @@ function latStr(lat: number): string {
 function longStr(long: number): string {
   return `${Math.abs(long)}° ${long >= 0 ? 'E' : 'W'}`;
 }
+
+export function compareFirestoreKeys(fieldA: string, fieldB: string): number {
+  return fieldA.localeCompare(fieldB);
```
yuchenshi, line 220: Please add a TODO comment like `// TODO: Use UTF-8 encoded byte order instead to match Firestore production behavior` with a link to docs: https://firebase.google.com/docs/firestore/manage-data/data-types

yuchenshi (approved):
Approved with one final nit on comments

yuchenshi:
> @yuchenshi Can I add the `Fix` word where I cite the issue in the PR description to automatically close it?

I've done that for you :)

src/components/Firestore/utils.ts
GabriFila: Done it

yuchenshi (approved):

yuchenshi:
Thanks for the contribution and thanks for working with me on this PR! I just merged this and it may take us a while to release the next version as we're going through a sizable change right now. I'll update this thread once we release, but no ETA yet. Cheers!

PR #474, opened 2021-01-07
Firestore sub-tabs behavior

AntonioRdzNav:
This is the **2nd PR** created to review the code of my intern project.

This PR focuses on the Firestore sub-tabs behavior: **Data and Requests**.

Since this PR breaks the original Firestore url **'/firestore'** into either **'/firestore/data'** or **'/firestore/requests'**, the Unit Tests of quite a few Firestore components broke. Therefore, this PR also focuses on fixing all those broken tests.

Finally, the routes of the **Requests Table** and **Request Details** render just a very basic UX mock file, I will add the proper UX and Unit Tests of those views on future PRs.


src/components/App/index.test.tsx
```diff
@@ -64,3 +65,37 @@ it('shows dialogs in the queue', async () => {
...
+  render(
+    <Provider store={store}>
+      <Router history={history}>
+        <App />
+      </Router>
+    </Provider>
+  );
+  await act(() => delay(100)); // Wait for tab indicator async DOM updates.
+  const { pathname, search } = history.location;
+  const redirectedCompleteUrl = pathname + search;
```
samhorlbeck, line 83: nit: eliminate the variable `redirectedCompleteUrl` and just put `pathname + search` into the expect statement

src/components/Firestore/Requests/RequestDetails/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+interface PropsFromParentComponent {
+  requestId?: string;
+}
+export type Props = PropsFromParentComponent;
```
samhorlbeck, line 22: Small nit: Avoid using `type`.

Instead, prefer

```ts
export interface Props extends PropsFromParentComponent {}
```

src/components/Firestore/Requests/RequestDetails/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const RequestDetails: React.FC<Props> = ({ requestId }) => (
+  <div data-testid="request-details">
+    <div> Request Details Header </div>
```
samhorlbeck, line 26: Nit: remove the spaces between the tags (`<div>`) and the content

src/components/Firestore/Requests/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    }}
+  >
+    <Switch>
+      <Route exact path="/firestore/requests">
+        <RequestsCard />
+      </Route>
+      <Route
+        exact
+        path="/firestore/requests/:requestId"
+        render={({ match }: any) => {
```
samhorlbeck, line 38: Is it possible to avoid `any` and give this an explicit type?

src/components/Firestore/index.scss
```diff
@@ -17,21 +17,45 @@
...
+.Firestore-sub-tabs {
+  margin-bottom: $firestore-sub-tabs-margin-bottom;
+  // shadow border at bottom
+  box-shadow: 0 -1px 0 rgba(0, 0, 0, 0.12) inset;
+
+  .Firestore-sub-tab-bar {
+    width: fit-content;
+    height: $firestore-sub-tabs-height;
+    .mdc-tab {
+      &:not(:first-of-type) {
```
samhorlbeck, line 43: Combine this with the `.mdc-tab` rule to avoid extra nesting

src/test_utils.tsx
```diff
@@ -163,3 +163,8 @@ export const wrapWithForm = <P, T, F = UseFormOptions<T>>(
 
   return { ...methods, triggerValidation, submit };
 };
+
+export function isTabActive(labelEl: HTMLElement): boolean {
+  const tabEl = labelEl.closest('.mdc-tab')!;
```
samhorlbeck, line 168: Remove the `!` assertion. On the next line, do:

```ts
return !!tabEl?.classList.contains('mdc-tab--active');
```

samhorlbeck (changes requested):
Nice work

yuchenshi (commented):
The routing changes are overkill -- it's much easier to keep the top level routing as-is `/firestore` and change only the nested `/data` or `/request` route. This should also reduce some tab-related changes

src/components/Firestore/index.tsx
```diff
@@ -84,10 +93,28 @@ export const Firestore: React.FC = React.memo(() => {
   const eject = useEjector();
 
   // TODO: do something better here!
-  const path = location.pathname.replace(/^\/firestore/, '');
+  const path = location.pathname.replace(/^\/firestore\/data/, '');
   const showCollectionShell = path.split('/').length < 2;
   const showDocumentShell = path.split('/').length < 3;
 
+  const subNavRoutes = routes.filter((r) => r.showInFirestoreSubNav);
```
yuchenshi, line 100: Please keep this local, no need to modify global `routes` array

src/index.tsx
```diff
@@ -39,6 +42,23 @@ const RouterWithInit = () => {
   useEffect(() => {
     store.dispatch(subscribeToConfig());
   }, []); // Empty-array means "run only once on init": https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
+
+  // Initialize global subscription to firestore requests evaluation
+  useEffect(() => {
```
yuchenshi, line 47: This won't work once websocket host and port is dynamic (derived from config). Better put this in sagas and subscribe in response to config fetch success action. Remember that ports may change (e.g. when emulators are restarted) so you'd need to unsubscribe and resubscribe if that happens.

src/components/Firestore/Requests/RequestDetails/index.tsx
AntonioRdzNav: I didn't know this approach was preferred for React Props, thank you!

src/index.tsx
AntonioRdzNav: We decided to not work on this code yet, since it will not be necessary until later on.

PR #476, opened 2021-01-08
Firestore request evaluations global subscription

AntonioRdzNav:
This is the **3rd content PR** created to review the code of my intern project.

This PR focuses on creating a **global subscription** to the websocket that receives request evaluations on real-time.

This solution is just temporal, since later the global subscription will have to be handled using **sagas** side-effects.

In addition to the global subscription, the code to simulate **admin** and **error** request evaluations was added by injecting mock data to incoming request evaluations. This code is useful for development purposes, but will be deleted once the server handles admin and error requests.


yuchenshi (approved):
Thanks for splitting it up! That was an easy approval

samhorlbeck (approved):

PR #478, opened 2021-01-13
PR-Firestore-Requests-Table-Row

AntonioRdzNav:
This is the **5th PR** created to review the code of my intern project.

This PR focuses on adding the **RequestsTableRow** content (UI, SCSS, and unit test).

Requests table with content:

image

Requests table on screen with smaller width:

image

As you may notice, the **request path column** is missing the **copy-icon button**, and is **truncating from the right**
 side (instead of from the left side), that is because I placed that content inside an independent component, I will add that component in a future PR.


src/components/Firestore/Requests/index.scss
```diff
@@ -0,0 +1,90 @@
...
+.Firestore-Request-Path {
+  position: relative;
+  // Display copy IconButton only when path is hovered
+  &:hover {
+    .Firestore-Request-Path-Container button {
+      visibility: visible;
+    }
+  }
+  .Firestore-Request-Path-Container {
+    // Flex behavior
```
yuchenshi, line 51:
```suggestion
```

src/components/Firestore/Requests/index.scss
```diff
@@ -0,0 +1,90 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    // Request path string
+    & > div {
```
yuchenshi, line 56: Use a descriptive class name instead

src/components/Firestore/Requests/types.ts
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CustomThemeType } from '../../../themes';
+import { RulesOutcome } from './rules_evaluation_result_model';
+
+export interface OutcomeData {
```
yuchenshi, line 20: These things are not worth their own interface / type definition since they are only used for the `outcomeData` const. Just let TypeScript infer the type there.

src/components/LogViewer/History.tsx
```diff
@@ -27,7 +27,7 @@ import { CompiledGetterCache } from './CompiledGetterCache';
 import { HighlightedJSON } from './HighlightedJSON';
 import { ParsedQuery, filtersToQueryString, isQueryMatch } from './QueryBar';
 
-const formatTimestamp = (timestamp: number) => {
+export const formatTimestamp = (timestamp: number) => {
```
yuchenshi, line 30: Please move this util function to a shared file

yuchenshi (approved):

src/utils.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Converts a timestamp into a consistent and formatted firebase-log date
+ */
+export const formatTimestamp = (timestamp: number) => {
```
yuchenshi, line 20: IMO we should just use `date.toLocaleTimeString('en-US', { hour12: false })` and remove this function altogether while we're at it.

yuchenshi (approved):

src/components/Firestore/Requests/RequestsCard/Table/TableRow.test.tsx
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright 2019 Google LLC
```
samhorlbeck, line 2: 2021

src/components/Firestore/Requests/utils.ts
```diff
@@ -14,6 +14,25 @@
  * limitations under the License.
  */
 
+import { OutcomeData } from './types';
+
+// Matches the material-icon name by request outcome
+const ICON_SELECTOR = {
```
samhorlbeck, line 20: Let's make this a `const enum`:

```ts
const enum OutcomeIcons {
  ALLOW = 'check_circle',
  // ...
};

// Then:
{icon: OutcomeIcons.ALLOW}
```

src/utils.tsx
samhorlbeck, line 20: If not going with Yuchen's suggestion, please export this as a function instead of a lambda variable

samhorlbeck: +1

samhorlbeck (changes requested):

src/utils.tsx
AntonioRdzNav: I agree, I will go with this solution instead!

src/components/Firestore/Requests/utils.ts
AntonioRdzNav: I think this constant should stay a dictionary, another component uses this dictionary to get the correct icon based on the outcome string, that code is just not part of this PR.

samhorlbeck: Ah okay, that's fine then

src/components/Firestore/Requests/index.scss
```diff
@@ -0,0 +1,84 @@
...
+  color: var(--mdc-theme-primary);
+  // Center icon
+  display: block;
+  margin: 0 auto;
+  text-align: center;
+}
+.Firestore-Request-Method {
+  text-transform: uppercase;
+  font-size: 13px;
+  font-family: Courier;
```
samhorlbeck, line 34: Do we have a fallback to "monospace" elsewhere or is it all just Courier (@yuchenshi )

samhorlbeck (approved):

src/components/Firestore/Requests/index.scss
AntonioRdzNav: I updated the font-family to **Courier** to match the style of the [Figma UX visual feedback](https://www.figma.com/file/oz3FtG70f3xfYVado4Qvha/Request-Viewer-Feedback?node-id=2%3A51) from marissa, but you folks let me know if I should change it.

image

yuchenshi: Let's be consistent with other existing monospace fonts in this project. Please take a look -- I believe `Roboto Mono` is used with some fallbacks

samhorlbeck: +1 to Yuchen's response. This is what I mean by fallback: https://stackoverflow.com/questions/4673963/css-fallback-fonts

AntonioRdzNav: I see, it makes sense. I took a look at other font fallbacks in the project and they indeed use `'Roboto Mono', 'monospace'`, so I will use: `font-family: Courier, 'Roboto Mono', monospace;`

AntonioRdzNav: After quick font discussion on gChat, we decided to get **rid of Courier** and just go with `font-family: 'Roboto Mono', monospace;` instead.

image

PR #488, opened 2021-01-21
PR-Firestore-Request-Details-Code-Viewer

AntonioRdzNav:
This is the **8th PR** created to review the code of my intern project.

This PR focuses on adding the **Code Viewer section** of the RequestDetails view.

In order to test the component that renders codemirror, I had to add a mocked DOM function that codemirror calls and jsdom does not handle (see **src/setupTests.js**)

Additionally, I restructured the **RequestDetails** folder by **splitting the files into 3 sub-folders** (Header, CodeViewer, InspectionSection) and an index file. I took this decision because the RequestDetails folder was getting very big.

Code Viewer for allowed request:

image

Code Viewer for denied request:

image

Code Viewer for error request:

image

Code Viewer for admin request:

image


src/components/Firestore/Requests/RequestDetails/CodeViewer/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+        successThemeBackground: successTheme.background,
+        errorThemePrimary: errorTheme.primary,
+        errorThemeBackground: errorTheme.background,
+      }}
+    >
+      <div
+        data-testid="request-details-code-viewer"
+        className="Firestore-Request-Details-Code-Section"
+      >
+        {!isAdminRequest && (
```
samhorlbeck, line 93: nit: extract this to a function that returns the correct element (either `CodeMirror` or `CodeViewerAdminRequest`)

src/components/Firestore/Requests/RequestDetails/CodeViewer/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  linesOutcome,
+  isAdminRequest,
+}) => {
+  const [
+    codeMirrorEditor,
+    setCodeMirrorEditor,
+  ] = useState<CodeMirror.Editor | null>(null);
+
+  // Creates HTMLElement of a Marker containing the outcome icon
+  const getMarkerElement = useCallback((outcome: RulesOutcome): HTMLElement => {
```
yuchenshi, line 48: If I read this correctly, this function doesn't access any variables in this component -- you may as well just move it to module scope and remove the `useCallback` wrapper (since it will be effectively constant).

src/components/Firestore/Requests/RequestDetails/CodeViewer/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    let marker = document.createElement('i');
+    marker.className = `CodeMirror-gutter-icon ${outcome} material-icons`;
+    marker.innerText = ICON_SELECTOR[outcome];
+    return marker;
+  }, []);
+
+  // Adds the corresponding Marker with the correct outcome icon
+  // to all lines that have an outcome (allow or deny)
+  const addLinesOutcomeGutters = useCallback(() => {
```
yuchenshi, line 57: Just inline this into the `useEffect` block below. No need for its own function or `useCallback`. In that way you can also remove the optional chaining.

src/components/Firestore/Requests/RequestDetails/CodeViewer/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    linesOutcome?.map(({ line, outcome }) => {
+      codeMirrorEditor?.setGutterMarker(
```
yuchenshi, line 59: FYI this won't remove old gutters if `lineOutcome` were to change.

yuchenshi (approved):

src/components/Firestore/Requests/RequestDetails/CodeViewer/index.tsx
samhorlbeck: This should be fine if the codemirror instance is destroyed and re-created when you go back and forth from the list / details view

src/components/Firestore/Requests/RequestDetails/CodeViewer/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const renderCodeViewer = () => {
+    if (isAdminRequest) {
+      return <CodeViewerAdminRequest />;
+    }
+    return (
+      <CodeMirror
+        value={firestoreRules}
+        options={{
+          theme: 'xq-light',
```
samhorlbeck, line 88: I would extract the code mirror options to a file-level constant up at the top of the file

samhorlbeck (approved):

src/components/Firestore/Requests/RequestDetails/CodeViewer/index.tsx
AntonioRdzNav: It is true, the codemirror instance is destroyed and re-created after navigating back and forth the requests table. Additionally, there is currently no way of updating the `linesOutcome` array of an existing request. Should we handle this case anyways? If not, I will at least at a TODO comment.

yuchenshi: A comment explaining why this is fine is good enough.

src/colors.tsx
```diff
@@ -16,6 +16,7 @@
 export const background = '#F6F7F9';
 export const primary = '#681DA8'; // Google Purple 900.
 export const secondary = 'rgba(0, 0, 0, .54)';
+export const ternary = 'rgba(0, 0, 0, .38)';
```
yuchenshi, line 19: 1. The contrast of this color is really bad -- is there a reason why we have to use this color for comments? I have concerns about accessibility. I don't see how this color can be safely used elsewhere either.

If we were to keep this: the name seems to imply it is like `primary` and `secondary` and may get confused with MDC theming. Thoughts on limiting it to component-scope for now and revisit?

yuchenshi (approved):

src/colors.tsx
AntonioRdzNav: I also see that I made a typo on the name of the constant, I meant to write `tertiary`, not `ternary`

I changed the color of comments to a light gray because Marissa suggested that change.

image

She actually suggested using the `textBlackSecondary`, but I changed it to `textBlackTertiary` because I thought the secondary text color did not have enough contrast with the color of the code.

image

Should we change the color? I see that the console uses a very contrasting red color for comments, we could maybe change it to red if we want to keep the color consistency.

image

AntonioRdzNav: Also, this color is used by the `InspectionSection` component in the next PR, and since it is an official firebase spec color, I think it could live as a constant in the `colors.tsx` file, changing the name to maybe `textBlackTertiary` could help avoiding any confusion.

yuchenshi: Let's keep it as `textBlackTertiary` for now and since we're using it in `InspectionSection` I'd say it can stay. We can revisit this later.

PR #489, opened 2021-01-21
PR-Firestore-Request-Details-Inspection-Section

AntonioRdzNav:
This is the **9th PR** created to review the code of my intern project.

This PR focuses on adding the **Inspection section** of the RequestDetails view.

image


src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <div className={`${INSPECT_BLOCK_CLASS}-Value`} title={value}>
+        {value}
+      </div>
+    );
+  };
+
+  return (
+    <ThemeProvider
+      options={{
+        inspectionMainBlockBackground: darkestGray,
```
yuchenshi, line 62: Don't extend the `--mdc-theme-` namespace to avoid confusion. Is there a color in material theme that is close enough to your desired gray? If not, just use a color value in CSS and avoid `ThemeProvider` (since this component won't respond to theming at all).

src/components/Firestore/Requests/RequestDetails/InspectionSection/index.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { InspectionElement } from '../../types';
+import InspectionSection from './index';
+
+describe('InspectionSection', () => {
+  const INSPECTION_MOCKED_DATA: InspectionElement[] = [
+    { label: '', value: '' },
```
yuchenshi, line 25: Please put some nonempty labels/values and test for those

src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+      }}
+    >
+      <div
+        className={classnames(
+          INSPECT_BLOCK_CLASS,
+          isMainBlock && MAIN_INSPECT_BLOCK_CLASS
+        )}
+        onClick={() => setIsExpanded(!isExpanded)}
+        role={isMainBlock ? 'inspection-main-block' : 'inspection-block'}
```
yuchenshi, line 71: Those are not valid HTML `role`s. Just simply test for the label here. No need for `getByRole`.

src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+      >
+        <span className={`${INSPECT_BLOCK_CLASS}-Label`}> {label} </span>
```
yuchenshi, line 73:
```suggestion
        <span className={`${INSPECT_BLOCK_CLASS}-Label`}>{label}</span>
```

Or did you mean to add spacing around the label? If so, use CSS for that.

yuchenshi (approved):

src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
AntonioRdzNav: The spaces were added automatically by prettier, thank you for noticing them!

src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
AntonioRdzNav: There is no existing color that is close enough to the desired gray, I actually came up with a solution using `ThemeProvider`, since the inspection-block can either be a main-block (which has a darker grey as background) or a regular-block (which uses the existing `grey100` as background).

Let me know what you think.

image
image

src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
AntonioRdzNav: I agree I should not use the HTML `role` in this case, and I did change the tests to get the element by text, however I also kept a `data-testid` (instead of the HTML `role`) because it is not only important to test that the element with such text exists, but also to test if it is a `main-block` or not

src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
yuchenshi: Sure, let's go with your latest proposed approach.

src/components/Firestore/Requests/RequestDetails/InspectionSection/InspectionBlock.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export const InspectionBlock: React.FC<Props> = ({
+  label,
+  value,
+  isMainBlock,
+  children,
+}) => {
+  // Main inspection-blocks start expanded by default
+  const [isExpanded, setIsExpanded] = useState<Boolean>(!!isMainBlock);
```
samhorlbeck, line 44: I believe this should be lower case `b` `boolean`

src/components/Firestore/Requests/RequestDetails/InspectionSection/index.tsx
```diff
@@ -33,7 +34,16 @@ const InspectionSection: React.FC<Props> = ({
     data-testid="request-details-inspection-section"
     className="Firestore-Request-Details-Inspection"
   >
-    Inspection Section
+    <InspectionBlock isMainBlock label="Query Information">
```
samhorlbeck, line 37: We generally try to use sentence case, so in this case the label should be "Query information"

src/components/Firestore/Requests/RequestDetails/InspectionSection/index.tsx
```diff
@@ -33,7 +34,16 @@ const InspectionSection: React.FC<Props> = ({
...
+      {inspectionQueryData?.map(({ label, value }) => (
+        <InspectionBlock key={label} label={label} value={value} />
+      ))}
+    </InspectionBlock>
+    <InspectionBlock isMainBlock label="Expressions Inspection">
```
samhorlbeck, line 42: Same here

samhorlbeck (approved):

PR #493, opened 2021-01-26
PR-Firestore-Request-Path

AntonioRdzNav:
This is the **10th PR** created to review the code of my intern project.

This PR focuses on adding the following components:

- **RequestPath**: responsible for `rendering the request-path`, `rendering the copy-icon-button`, `truncating the request-path` (from the left side) and `copying to clipboard` the request-path after the copy-icon-button is clicked.
- **CopyPathNotification**: responsible for `rendering the Snackbar` (notification) that appears after the copy-icon-button is clicked.

Requests table at screen with big-size width (approx 1250px):

image

Request table at screen with small-size width (650px):

image

Request table in mobile (Pixel 2):

image

Request details at screen with big-size width (approx 1250px):

image

Request details at screen with small-size width (650px):

image

Request details in mobile (Pixel 2):

image

Copy-path-notification in requests table:

image

Copy-path-notification in request details:

image

Diagram to explain the relationship between **RequestPath** and **CopyPathNotification**:

image


src/components/Firestore/Requests/RequestPath/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+// Copies path without spaces to clipboard
+// and triggers copy notification (SnackBar)
+function copyPathToClipboard(
+  resourcePath: string,
+  setShowCopyNotification: (value: boolean) => void
+) {
+  navigator.clipboard
+    .writeText(resourcePath.replace(/\s/g, ''))
+    .then(() => setShowCopyNotification(true))
+    .catch(() => setShowCopyNotification(false));
```
yuchenshi, line 32: Consider showing an error when this throws (or if `navigator.clipboard` is unsupported)

src/components/Firestore/Requests/utils.ts
```diff
@@ -34,15 +37,43 @@ export const OUTCOME_DATA: OutcomeDataPicker = {
...
+  useEffect(() => {
+    window?.addEventListener('resize', debouncedHandleWindowResize);
+    return () =>
+      window?.removeEventListener('resize', debouncedHandleWindowResize);
+  }, [debouncedHandleWindowResize]);
+
+  // Set the initial width of path-container
+  // NOTE: this code only executes the first time the component renders
+  useEffect(() => {
+    setPathContainerWidth(getPathContainerWidth());
```
yuchenshi, line 75: I think you can safely combine this into the `useEffect` above (maybe as the first thing)

yuchenshi (approved):

src/components/Firestore/Requests/RequestPath/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    pathHtmlElement.textContent = newTruncatedPathString;
+  }
+}
+
+interface Props {
+  resourcePath: string;
+  setShowCopyNotification: (value: boolean) => void;
+  requestPathContainerWidth?: number;
+  // Props used for testing purposes only
+  mockedPathOffsetWidth?: number;
```
samhorlbeck, line 103: I'm not super excited about adding a couple of properties whose only use is testing. See my other comment on the sizing function

src/components/Firestore/Requests/RequestPath/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    .catch(() => setShowCopyNotification(false));
+}
+
+// Function that truncates the request path from the left side of the string.
+// It first calculates the available width of the pathHtmlElement, and how many
+// characters fit inside the width. Based on that, it calculates a new truncated string.
+// For optimization purposes, the truncation is always calculated, but the HTMLElement
+// is only updated when the new truncated string is different than the HTMLElement's textContent.
+function truncateRequestPathFromLeft(
+  pathTextElement: React.RefObject<HTMLDivElement>,
```
samhorlbeck, line 41: I'm not too excited about having properties only used for testing. Here's one suggestion to avoid this:

1. Create a new (exported for testing) function that accepts `pathHtmlElement` directly as an `HTMLDivElement`.
2. Have this function pull out the `.current` value and just call the new function
3. Add some tests that directly call this new function. It won't go through the DOM, but that's okay in this case
    1. Now, you can pass in a plain objects with `textContent` and `offsetWidth` properties. (You'll need to cast them with `as unknown as HTMLDivElement` or `as unknown as HTMLButtonElement`)
    2. After the function call, you can check the text was truncated correctly by examining the `textContent` field on this mock object

Your test will no longer directly check the interaction with the DOM, but I think that's worth it in this case to avoid the hack

src/components/Firestore/Requests/RequestPath/index.tsx
AntonioRdzNav: Can I use the error-reason that comes from the rejected Promise? or would it be more helpful to display a custom string like: `'Could not copy request-path to clipboard` ?

image

src/components/Firestore/Requests/RequestPath/index.tsx
AntonioRdzNav: I added one solution from what we dicussed yesterday in gChat:

- Now the `offsetWidth` properties of the **request-path** and the **copy-icon-button** are mocked inside the test file, therefore, the UI is not modified at all (no need to inject test-props).
- With this approach, the component can be rendered in the DOM.

Let me know what you think.

src/components/Firestore/Requests/CopyPathNotification/index.scss
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.Firestore-Requests-Copy-Path-Snackbar {
+  & .mdc-snackbar__surface {
+    background-color: #051e34; // navy800
```
samhorlbeck, line 19: nit: two spaces between code and inline comment

src/components/Firestore/Requests/RequestPath/index.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const MOCKED_PATH =
+    '/collection1/collection1_ID/subcollection/subcollection_ID';
+  const COPY_ICON_TEXT = 'content_copy';
+
+  describe('RequestPath => truncation', () => {
+    // Manually set a custom offsetWidth value of a given HTML element
+    // NOTE: using this mocked values is important because jsdom does not handle
+    // layouts (which means that every variable that contains a size measurement
+    // will be equal to 0 during a test)
+    function setHTMLElementOffsetWidth(
```
samhorlbeck, line 32: Move this function to the top-level of the file since it doesn't access any of the enclosing state

src/components/Firestore/Requests/RequestPath/index.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+    it('renders truncated request path when complete path does not fit in width of container', async () => {
+      // Prop of width of container is not yet given
+      // (simulates the behavior of the width not yet been calculated)
+      const { getByText, getByTestId, rerender } = render(
+        truncationJSXContent()
+      );
+      const pathHtmlElement = getByTestId('request-path-text');
+      const copyIconButtonHtmlElement = getByText(COPY_ICON_TEXT);
+      setHTMLElementOffsetWidth(pathHtmlElement, 250);
```
samhorlbeck, line 79: Yes this looks way cleaner, thanks again :+1:

samhorlbeck (approved):

PR #495, opened 2021-02-03
Tl/e2e

tjlav5:


yuchenshi:
1. Would you mind briefly describe why you arrived at `puppeteer` instead of the alternatives in the PR description?
2. Open question: Is there anyway we can see the visual differences on CI if tests fail?

PR #510, opened 2021-02-19
Format all supported files with prettier + eslint.

yuchenshi:
Most interesting file is `package.json` and `nodejs.yml` (for running it on GitHub CI). `.prettierignore` is also worth a look. All other changes are auto-generated and should be rubber-stamped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,8 +74,10 @@
     "test:debug": "react-scripts --inspect-brk test --env=./custom-jest-environment.js --runInBand",
     "test:coverage": "react-scripts test --coverage --watchAll=false --env=./custom-jest-environment.js",
     "eject": "react-scripts eject",
-    "lint": "eslint 'src/**/*.{ts,tsx}'",
-    "fmt": "prettier --write 'src/**'"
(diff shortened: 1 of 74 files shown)


package.json
```diff
@@ -133,9 +135,9 @@
     }
   },
   "lint-staged": {
-    "src/**/*.{js,jsx,ts,tsx,json,css,md,scss}": [
-      "prettier --write",
-      "git add"
+    "*.{ts,tsx}": [
+      "eslint --fix",
+      "prettier --write"
```
NothingEverHappens, line 140: What lint -staged? Is this for CI? Do we need to write here?

NothingEverHappens (approved):

package.json
yuchenshi: This is the pre-commit hook, which can fix some eslint issues and format issues automatically when you run `git commit`. So yeah, `--write` is needed to actually fix that file. (FYI the CI check runs `npm lint`, which runs `prettier --check` instead of `--write`.)

PR #521, opened 2021-02-25
Update README & LICENSE

PawelBorkar:
Add hyperlink to the License in README and Copyrights to LICENSE.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2019-present Google LLC. All rights reserved.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2019-present Google LLC. All rights reserved.
+   Copyright [yyyy] [name of copyright owner]
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
```
yuchenshi, line 190: This is part of the original license text that serves as an appendix on how to apply the license to the readers own project and should not be changed.

README.md
```diff
@@ -101,4 +101,6 @@ the server as described above.
 
 ## License
 
-Apache 2.0
+[Apache-2.0 License](https://github.com/firebase/firebase-tools-ui/blob/master/LICENSE)
+<br>
+Copyright 2019-present Google LLC. All rights reserved.
```
yuchenshi, line 106:
```suggestion
Copyright 2019-2021 Google LLC. All rights reserved.
```

yuchenshi (changes requested):
Thanks for the PR. See comments below.

LICENSE
PawelBorkar: Should I add a copyright header then instead of adding in here as it was done in the other public repositories like tensorflow?

README.md
PawelBorkar: Done as suggested.

LICENSE
yuchenshi: No, the Apache License is to be kept verbatim. Yes, a few files were missing a copyright header, but I'll do a follow up PR myself using some automated tools. It's not really worth a human's time to go in and fix each one by hand.

Please revert this file and I'll merge the README changes below.

PawelBorkar: Done as suggested. Thanks for the clarifications.

yuchenshi (approved):

PR #524, opened 2021-03-03
Storage Emulator

NothingEverHappens:


yuchenshi:
Please kindly amend the commit with your `@google.com` email address and the CLA warning will go away. Thanks!

src/components/Storage/Canvas/BucketPicker/BucketPicker.test.tsx
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright 2019 Google LLC
```
yuchenshi, line 2: New files should have `2021` as the copyright year. (I see a few 2019s and 2020s throughout new files on this PR and you may want to do a global find and replace for the changed files)

src/components/common/_material-overrides.scss
```diff
@@ -91,3 +91,7 @@
     @include elevation.elevation(9);
   }
 }
+
+.material-icons {
+  font-family: 'Material Icons Extended';
```
yuchenshi, line 96: Is `'Material Icons Extended'` a superset? If so, can / shall we remove the Material Icons CSS from `public/index.html`?

src/firebase.ts
```diff
@@ -57,10 +59,7 @@ export function useEmulatedFirebaseApp(name: string, config: any) {
   const projectId = useProjectId();
 
   const app = useMemo(() => {
-    const app = firebase.initializeApp(
-      { ...config, projectId },
-      `${name} component::${Math.random()}`
-    );
+    const app = firebase.initializeApp({ ...config, projectId });
```
yuchenshi, line 62: Please use a unique app ID or this may throw upon emulator restart (`config` and `projectId` may change / flicker). I'd suggest reverting this change.

src/setupTests.js
```diff
@@ -36,3 +36,6 @@ Object.defineProperty(window, 'scrollTo', {
 // own label.
 base.randomId = (prefix) =>
   `${prefix}-${(Math.random() + Math.random() + 1).toString(36).substring(2)}`;
+
+// Storage tests take over 5 seconds
+jest.setTimeout(10000);
```
yuchenshi, line 41: Which tests take >5 seconds? If there are only a few, just change the timeout for those tests or test suites.

src/components/common/_material-overrides.scss
NothingEverHappens: it's weird, the font family has to be `Material Icons Extended`, but both imports must be present, otherwise it stops showing regular icons.

src/setupTests.js
NothingEverHappens: Pretty much all storage tests that interact with the back-end.
I could potentially go and set it one by one if you think it's better?

yuchenshi: That's unfortunate -- let's set the timeout on the storage tests suites (for whole `describe` blocks).

custom-jest-environment.js
```diff
@@ -22,19 +22,19 @@ const JsdomEnvironment = require('jest-environment-jsdom');
 class EmulatorUiTestEnvironment extends JsdomEnvironment {
   constructor(config) {
     super(
-      Object.assign({}, config, {
```
yuchenshi, line 25: revert this (whitespaces only?)

src/setupTests.js
NothingEverHappens: Ok, once I connected the actual emulator, the tests became fast, so I dropped this line.

src/components/Storage/api/useCreateFolder.tsx
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import { useBucket } from './useBucket';
+import { useStorageConfig } from './useStorageConfig';
+
+export function useCreateFolder() {
+  const config = useStorageConfig();
+  const [bucket] = useBucket();
+
+  const data = `--boundary
```
yuchenshi, line 24: Extract this to module-level constant

src/components/Storage/api/useCreateFolder.tsx
```diff
@@ -0,0 +1,54 @@
...
+Content-Type: application/json
+
+{"contentType":"text/plain"}
+--boundary
+Content-Type: text/plain
+
+--boundary--`;
+
+  async function createFolder(fullPath: string) {
+    function normalizePath(fullPath: string) {
```
yuchenshi, line 34: Just inline this (`const encodedPath = ...`). This isn't just normalization -- it also encodes

src/components/Storage/api/usePath.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    const [path, setPath] = result.current;
+    return {
+      path,
+      setPath,
+      history,
+    };
+  }
+
+  it('picks up bucket name from the URL', async () => {
```
yuchenshi, line 53:
```suggestion
  it('picks up path from the URL', async () => {
```

src/components/Storage/api/usePath.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { StorageRouterParams } from '../types';
+import { useBucket } from './useBucket';
+
+export function usePath(): [string, (bucket: string) => void] {
+  const { path } = useParams<StorageRouterParams>();
+  const [bucket] = useBucket();
+  const history = useHistory();
+
+  function setPath(path: string) {
+    while (path.includes('//')) {
```
yuchenshi, line 29: This should just use regex for global replace `const MULTI_SLASH_GLOBAL = new RegExp('//+', 'g')`

src/components/Storage/api/useSelectedFile.tsx
```diff
@@ -0,0 +1,17 @@
+import { useState } from 'react';
+
+import { StorageFile, StorageItem } from '../types';
+
+export function useSelectedFile(
```
yuchenshi, line 5: This hook is overkill. We just need to keep `fullPath` in state at its only call site `StorageCard`.

src/components/Storage/api/useStorageConfig.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+describe('useStorageConfig', () => {
+  async function setup() {
+    const Wrapper: React.FC = ({ children }) => {
+      return <FakeStorageStoreProvider>{children}</FakeStorageStoreProvider>;
+    };
+    return renderHook(() => useStorageConfig(), { wrapper: Wrapper }).result;
+  }
+
+  it('picks up bucket name from the URL', async () => {
```
yuchenshi, line 31:
```suggestion
  it('returns host and port from provider', async () => {
```

src/components/Storage/api/useStorageConfig.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { EmulatorConfig } from '../../../store/config';
+import { getStorageConfigResult } from '../../../store/config/selectors';
+import { hasError } from '../../../store/utils';
+
+// Suspense aware wrapper.
+export function useStorageConfig(): EmulatorConfig {
+  const config = useSelector(getStorageConfigResult);
+
+  if (!config) {
+    throw Promise.resolve();
```
yuchenshi, line 28: This will probably just cause repeated rerenders until storage config is loaded. Worse, if you are on the storage tab and then storage emulator stops (e.g. across restart), this even have potential to go in infinite loop.

While I'm also eager to see this solved, please wait until we have go/fire-emulator-concurrent before we jump into this. For this PR, please just copy the pattern for e.g. Firestore w.r.t config.

yuchenshi (changes requested):

src/components/Storage/api/useStorageConfig.tsx
yuchenshi: Just to clarify, there's no need to rewrite the whole thing -- we just need to remove suspension from this hook and just get the fields from `useSelector`. Nested child components can still use this hook (and return type can be changed to not-nullable since we already got that deep in). Just need to change the route component (top level) a bit to resemble Firestore for loading / error states.

src/components/Storage/api/useSelectedFile.tsx
yuchenshi: Sounds reasonable to me. I'd still suggest keeping only `fullPath` and not the full object in state, but that is lower priority.

src/components/Storage/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const StorageRouteWrapper: React.FC = ({ children }) => {
+  const configResult = useSelector(getStorageConfigResult);
+
+  return handle(configResult, {
+    onNone: () => <Spinner span={12} message="Storage Emulator Loading..." />,
+    onError: () => <StorageRouteDisabled />,
+    onData: (config) =>
+      config === undefined ? <StorageRouteDisabled /> : children,
+  }) as any;
```
yuchenshi, line 81: nit: provide a type argument to `handle`, don't cast to any

src/components/Storage/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { useBuckets } from './api/useBuckets';
+import { StorageCanvas } from './Canvas/Canvas';
+import { StorageCard } from './Card/StorageCard';
+import { storagePath } from './common/constants';
+import styles from './index.module.scss';
+import { StorageFirebaseAppProvider } from './providers/FirebaseAppProvider';
+
+export const Storage: React.FC = () => {
+  return (
+    <Suspense fallback={<Spinner message="Loading storage..." />}>
```
yuchenshi, line 35: nit: move this tag to be inside `StorageRouteWrapper` (nothing to suspend in `StorageRouteWrapper` now that we've moved back to redux)

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useState } from 'react';
+
+function useSelected<T>(options: T[]): [Set<T>, (t: Set<T>) => void] {
+  const [selectedValue, setSelected] = useState<Set<T>>(new Set());
```
yuchenshi, line 20:
```suggestion
  const [selectedValue, setSelected] = useState(new Set<T>());
```

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+        selected.delete(item);
+      }
+    }
+
+    setSelected(new Set([...selected]));
+  }
+
+  function addToStack(item: T) {
+    stack.push(item);
+    setStack([...stack]);
```
yuchenshi, line 56: Don't mutate existing arrays, just do `setStack([...stack, item])`

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  return {
+    allSelected: selected.size === options.length && selected.size > 0,
+    allIndeterminate: selected.size > 0 && selected.size < options.length,
+    isSelected(item: T) {
+      return selected.has(item);
+    },
+    selected,
+    toggleSingle(item: T, checked?: boolean) {
+      if (!checked ?? selected.has(item)) {
```
yuchenshi, line 67: I don't think this does what you want -- `!undefined` will just give you `true` and `??` never gets triggered. Better way to handle this is adding a default value to the param itself (one line above)

src/components/Storage/Card/Table/Table.tsx
```diff
@@ -0,0 +1,207 @@
...
+      selected={selectedFile === item}
+      className={styles.row}
+    >
+      <DataTableCell hasFormControl>
+        <Checkbox
+          aria-labelledby={fileNameId}
+          checked={selection.isSelected(item.fullPath)}
+          onClick={(event: MouseEvent<HTMLInputElement>) => {
+            event.stopPropagation();
+            return selection.toggleSingle(
```
yuchenshi, line 164: Call `toggleContinuous` if `Shift` is pressed for real. We also need a test for this case

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+        toggleAllItems([item], false);
+        setStack(stack.filter((i) => i !== item));
+      } else {
+        toggleAllItems([item], true);
+        addToStack(item);
+      }
+    },
+    toggleContinuous(item: T) {
```
yuchenshi, line 75: Also needs a `checked` argument

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+      return selected.has(item);
+    },
+    selected,
+    toggleSingle(item: T, checked?: boolean) {
+      if (!checked ?? selected.has(item)) {
+        toggleAllItems([item], false);
+        setStack(stack.filter((i) => i !== item));
+      } else {
+        toggleAllItems([item], true);
+        addToStack(item);
```
yuchenshi, line 72: Should this replace the stack? (i.e. clear the previous stack and put `item` as the sole item)

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+  return [
+    new Set(options.filter((path) => selectedValue.has(path))),
+    setSelected,
+  ];
+}
+
+// Takes a list of elements and allow to select multiple like in gmail.
+export function useMultiselect<T>(options: T[]) {
+  const [selected, setSelected] = useSelected(options);
+  const [stack, setStack] = useState<T[]>([]);
```
yuchenshi, line 31: I'm not sure if I completely understood the "stack" concept here, so please let me know if I'm mistaken in the comments below. Would appreciate some comments on just storing the last clicked item isn't enough -- and may be a test case too.

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+    allSelected: selected.size === options.length && selected.size > 0,
+    allIndeterminate: selected.size > 0 && selected.size < options.length,
+    isSelected(item: T) {
+      return selected.has(item);
+    },
+    selected,
+    toggleSingle(item: T, checked?: boolean) {
+      if (!checked ?? selected.has(item)) {
+        toggleAllItems([item], false);
+        setStack(stack.filter((i) => i !== item));
```
yuchenshi, line 69: What does this do? Why do we only remove it when unchecking?

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+        this.toggleSingle(item);
+      }
+
+      const between = findBetween(stack[stack.length - 1], item);
+
+      toggleAllItems(between, !this.isSelected(item));
+      addToStack(item);
+    },
+    clearAll() {
+      selected.clear();
```
yuchenshi, line 86: Do we need to mutate `selected` here? Why can't we just `setSelected(new Set())`?

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+      setSelected(new Set([...selected]));
+    },
+    toggleAll() {
+      if (selected.size === options.length) {
+        selected.clear();
```
yuchenshi, line 91: same

src/components/Storage/common/useMultiselect.ts
yuchenshi: Also, shall we clear the stack?

src/components/Storage/common/useMultiselect.ts
```diff
@@ -0,0 +1,100 @@
...
+      if (!checked ?? selected.has(item)) {
+        toggleAllItems([item], false);
+        setStack(stack.filter((i) => i !== item));
+      } else {
+        toggleAllItems([item], true);
+        addToStack(item);
+      }
+    },
+    toggleContinuous(item: T) {
+      if (selected.size === 0) {
```
yuchenshi, line 76: Did you mean `stack.length === 0`? If not, please handle the case where `stack` is empty

yuchenshi (commented):
All is good, except continuous selection (Shift+click). This isn't actually enabled in code, and I don't fully grasp the logic there. To reduce risk, consider deferring this feature and file a follow-up bug instead of addressing those comments below. Other comments should still be addressed.

src/components/Storage/Card/Table/Table.tsx
yuchenshi: Or consider deferring this

NothingEverHappens: Will do this as a follow up

src/components/Storage/common/useMultiselect.ts
NothingEverHappens: I actually dropped second argument, I think the source of truth should be in the hook.
I also pinged you in the chat if you want to discuss that offline

src/components/Storage/common/useMultiselect.ts
NothingEverHappens: I'll add a comment in the code

src/components/Storage/common/useMultiselect.ts
NothingEverHappens: added a comment

src/components/Storage/Card/Table/Table.tsx
```diff
@@ -0,0 +1,221 @@
...
+      onClick={navigate}
+      selected={selectedFile === item}
+      className={styles.row}
+    >
+      <DataTableCell hasFormControl>
+        <Checkbox
+          aria-labelledby={fileNameId}
+          checked={selection.isSelected(item.fullPath)}
+          onChange={(event: MouseEvent<HTMLInputElement>) => {
+            event.stopPropagation();
```
yuchenshi, line 174: If I'm not mistaken, click events will now bubble to `navigate` (e.g. checking a folder may navigate to it instead) -- consider moving this to a separate `onClick` handler. Please test manually if necessary.

src/components/Storage/api/useStorageFiles.tsx
```diff
@@ -0,0 +1,207 @@
...
+  return {
+    files,
+    setBucket,
+    bucket,
+    path,
+    setPath,
+    bucketHasAnyFiles: bucketHasAnyFiles.data,
+    getLocation,
+    async uploadFiles(uploadedFiles: File[], folderName?: string) {
+      // Optimistically display new result on top with the loading state.
```
yuchenshi, line 165:
```suggestion
      // Optimistically display new file on top with the loading state.
```

src/components/Storage/api/useStorageFiles.tsx
```diff
@@ -0,0 +1,207 @@
...
+      await bucketHasAnyFiles.mutate();
+    },
+    openAllFiles,
+    async createFolder(name: string) {
+      createFolder(path + '/' + name);
+      await result.mutate();
+      await bucketHasAnyFiles.mutate();
+    },
+    async deleteFiles(paths: string[]) {
+      // Optimistically hide result
```
yuchenshi, line 193:
```suggestion
      // Optimistically hide file
```

src/components/Storage/common/useMultiselect.test.ts
```diff
@@ -0,0 +1,234 @@
...
+import { act, renderHook } from '@testing-library/react-hooks';
+
+import { useMultiselect } from './useMultiselect';
+
+describe('useMultiSelect', () => {
+  const item1 = 'a';
+  const item2 = 'b';
+
+  function setup() {
+    const items = [item1, item2, 'c', 'd', 'e'];
```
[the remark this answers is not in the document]
yuchenshi: This isn't addressed

src/components/Storage/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const configResult = useSelector(getStorageConfigResult);
+
+  return handle(configResult, {
+    onNone: () => <Spinner span={12} message="Storage Emulator Loading..." />,
+    onError: () => <StorageRouteDisabled />,
+    onData: (config) =>
+      config === undefined ? (
+        <StorageRouteDisabled />
+      ) : (
+        <Suspense fallback={'lol'}>
```
yuchenshi, line 43:
```suggestion
        <Suspense fallback={<Spinner span={12} message="Storage Emulator Loading..." />}>
```

This is application code (user visible) so let's put some nicer stuff

src/components/Storage/Card/Table/Header/StorageHeader/UploadButton/UploadButton.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const file = new File([''], 'pirojok.txt');
+
+    const upload = getByLabelText('Upload');
+
+    await act(async () => {
+      fireEvent.change(upload, { target: { files: [file] } });
+    });
+
+    expect(await waitForNFiles(1)).toBe(true);
+  }, 20000);
```
yuchenshi, line 37: nit: Does this test really take that long?

yuchenshi (approved):
Approved with nits

src/components/Storage/Card/Table/Table.tsx
NothingEverHappens: Oh, wow, good call

yuchenshi (approved):
All is good, just need to pull in the SDK and make tests pass

src/components/Storage/Card/Table/Header/ActionHeader/ActionHeader.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useStorageFiles } from '../../../../api/useStorageFiles';
+import { UseMultiselectResult } from '../../../../common/useMultiselect';
+import { StorageItem } from '../../../../types';
+import styles from './ActionHeader.module.scss';
+
+interface ActionHeaderProps {
+  selection: UseMultiselectResult;
+}
+
+function useAreFoldersSelected(files: StorageItem[], selection: Set<string>) {
```
yuchenshi, line 33: nit: this is not a hook and please remove the `use` prefix to reduce confusion. (It's a stateless pure function whose return value only varies based on input arguments)

yuchenshi (approved):

src/components/Storage/Card/Table/Header/ActionHeader/ActionHeader.tsx
kirjs: Ah, good call. done

src/components/Storage/testing/StorageStoreProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { AppState } from '../../../store';
+
+export const StorageStoreProvider: React.FC = ({ children }) => {
+  const projectId = '';
+  const hostAndPort = process.env.FIREBASE_STORAGE_EMULATOR_HOST;
+
+  if (!hostAndPort) {
+    throw new Error(
+      'Provide storage emulator address by setting FIREBASE_STORAGE_EMULATOR_ADDRESS'
```
yuchenshi, line 29:
```suggestion
      'Provide storage emulator address by setting FIREBASE_STORAGE_EMULATOR_HOST'
```

PR #572, opened 2021-05-25
Support recursive deletes for document.

yuchenshi:
Screen Shot 2021-05-24 at 6 31 27 PM

Tests won't pass until https://github.com/firebase/firebase-tools/pull/3402 is released, but code should be ready for review. Want to get this approved before I address collection deletes.


src/components/Firestore/dialogs/DeleteDocumentDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+              label="Also delete nested collections and documents"
+              checked={recursive}
+              onChange={(event) => {
+                setRecursive(event.currentTarget.checked);
+              }}
+            />
+          </div>
+          {recursive && (
+            <Callout aside type="note">
+              Cloud Functions triggers WON'T be executed.
```
samtstern, line 81: nit: maybe `<b>will not</b>` instead of `WON'T`?

samtstern (approved):

src/components/Firestore/Document.test.tsx
```diff
@@ -180,3 +186,153 @@ it('shows the selected document-collection when there are collection and documen
...
+    deleteDocument.click();
+    return waitForDialogsToOpen();
+  });
+
+  const recursiveCheckbox = getByRole('checkbox', {
+    name: 'Also delete nested collections and documents',
+  });
+
+  await act(async () => {
+    fireEvent.click(recursiveCheckbox);
```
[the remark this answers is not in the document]
yuchenshi: Updated these to use `.click()`. There's no need to use `fireEvent` in this case, but can be used for other events without handy trigger methods built-in to DOM.

src/components/Firestore/dialogs/DeleteDocumentDialog.tsx
yuchenshi: Done.

PR #580, opened 2021-05-26
Fix Firestore sometimes not using emulator and app leaks.

yuchenshi:
## Problem

Sometimes when Emulator Suite is restarted while on the Firestore tab, `useEmulator()` is sometimes not correctly called on the Firestore instance. It's not easy to reproduce, but can be done by killing `firebase emulators:start` with `Ctrl-C` and run the command again, and again. (In my experience, it depends on the polling timing and usually happens around the second or third time).

## The fix

I've tried to fix this in various ways, but the only way I can be sure of is to move the `useEmulator` call right next to the app initialization. That's also the recommended timing anyway and we don't need to reason about async ordering. This also fixes the leaking of app instances (i.e. more and more apps will keep being created and never deleted without the fix).

Firestore SDK lazy loading has been disabled for now. Sorry, but that just makes debugging why such things happen much harder and may be part of the cause. Feel free to investigate and add it back in the future with adequate testing.


src/firebase.ts
```diff
@@ -65,23 +66,33 @@ export const getStorageApp = (host: string, port: string) => {
...
 
-export function useEmulatedFirebaseApp(name: string, config: any) {
+export function useEmulatedFirebaseApp(
+  name: string,
+  config: any,
+  initialize?: (app: firebase.app.App) => void
+) {
   const projectId = useProjectId();
-
-  const app = useMemo(() => {
```
yuchenshi, line 71: Hooks are called during render phase and should not have side effects like this. Wrapping side effects in `useMemo` won't help. The right way is to `useEffect` for side effects and pair them up with cleanups.

To see the total active apps, run `firestore.app.firebase_.apps` in console. You'll see twice as many apps being created than deleted without the fix.

src/components/Firestore/index.tsx
```diff
@@ -88,11 +81,6 @@ export const Firestore: React.FC = React.memo(() => {
   const showCollectionShell = path.split('/').length < 2;
   const showDocumentShell = path.split('/').length < 3;
 
-  useEffect(() => {
```
yuchenshi, line 91: This is a duplicate of `FirestoreDevTools`.

yuchenshi:
DO NOT MERGE: Just noticed that this causes the app to be recreated on every router navigation. Investigating.

src/firebase.ts
```diff
@@ -65,23 +66,49 @@ export const getStorageApp = (host: string, port: string) => {
   return app;
 };
 
-export function useEmulatedFirebaseApp(name: string, config: any) {
+/**
+ * Get an JS SDK App instance with emulator Admin auth enabled.
```
sam-gc, line 70: nit "a" not "an"

sam-gc (approved):

NothingEverHappens (approved):

src/firebase.ts
yuchenshi:
```suggestion
 * Get a JS SDK App instance with emulator Admin auth enabled.
```

PR #581, opened 2021-05-28
Add swr-based emulator config hooks.

yuchenshi:
Progress and plan:

- [x] Add new providers and hooks
- [x] Migrate Home
- [x] Migrate AppDisconnected 
- [x] Migrate Firestore
- [x] Migrate Storage
- [ ] Migrate RTDB (and migrate databases list to `useSwr`, removing redux parts)
- [ ] Migrate Auth (keep Redux locally at least for now)
- [ ] Delete tons of boilerplate for `store/config` + more

----

Note to reviewers:

Start from `src/components/common/EmulatorConfigProvider.tsx` where new stuff is introduced. See `Home` and `AppDisconnected` to see them in action. Then move on to bigger changes like Firestore or Storage. Per-commit view on GitHub can be handy -- for example Storage changes are isolated in their own commit.

PR #590, opened 2021-06-08
Migrate Database and LogViewer off Redux

yuchenshi:
Follow up to #581. Only Auth is left.


src/components/Database/DatabaseContainer.tsx
```diff
@@ -113,12 +93,39 @@ export const DatabaseContainer: React.FC<Props> = ({
...
+  );
+  return <DatabaseContainer databases={data} {...props} />;
+};
+
+export interface DatabaseInfo {
+  name: string;
+}
+
+async function fetchDatabases(
+  _: unknown,
```
sam-gc, line 115: Why is the first parameter like this?

sam-gc (approved):

PR #591, opened 2021-06-08
Nuke config from Redux and update Auth.

yuchenshi:
This PR removes emulator config from Redux. Auth state is still in Redux + Sagas, but it now receives config from the hook instead of the store. As a side benefit, this should fix http://b/171739543.

This is the last PR in the series and there is no immediate plan to migrate Auth to anything else.


src/store/auth/sagas.ts
```diff
@@ -56,12 +60,36 @@ export function* fetchAuthUsers(): any {
...
-  if ('data' in result && result.data) {
-    yield put(authFetchUsersRequest());
-    yield put(getAllowDuplicateEmailsRequest());
+export const AUTH_API_CONTEXT = 'AUTH_API_CONTEXT';
+
+export function* initAuth({ payload }: ActionType<typeof updateAuthConfig>) {
+  if (!payload) {
+    return;
   }
+  yield setContext({
```
yuchenshi, line 69: Context is only available to this sagas and its children. No leaking :)

src/store/auth/sagas.ts
```diff
@@ -56,12 +60,36 @@ export function* fetchAuthUsers(): any {
...
+    takeLatest(getType(setUserDisabledRequest), setUserDisabled),
+    takeLatest(
+      getType(setAllowDuplicateEmailsRequest),
+      setAllowDuplicateEmails
+    ),
+    takeLatest(
+      getType(getAllowDuplicateEmailsRequest),
+      getAllowDuplicateEmails
+    ),
+    put(authFetchUsersRequest()),
```
yuchenshi, line 90: Moved here so that each config change (emulator restarting) triggers a fetch

NothingEverHappens (approved):

PR #619, opened 2021-07-30
Test switching Storage JS SDK to browser build.

yuchenshi:


yuchenshi:
ESM build doesn't work, had to use the CJS build in #611

PR #637, opened 2021-08-29
Fix windows issue with npm start

### LINKED ISSUE #210: SassError: Can't find stylesheet to import "@material/theme/mixins";
> Failed to compile
> ./src/index.scss (./node_modules/css-loader/dist/cjs.js??ref--6-oneOf-5-1!./node_modules/postcss-loader/src??postcss!./node_modules/resolve-url-loader??ref--6-oneOf-5-3!./node_modules/sass-loader/dist/cjs.js??ref--6-oneOf-5-4!./src/index.scss)
> SassError: Can't find stylesheet to import.
>    ╷
> 24 │ @.use "@material/theme/mixins";
>    │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>    ╵
>   node_modules\@material\card\_variables.scss 24:1     @.forward
>   node_modules\@material\card\_index.scss 1:1          @.use
>   src\components\common\_material-overrides.scss 18:1  @.import
>   C:\Users\$USER$\firebase-tools-ui\src\index.scss 24:9                                           root stylesheet

npm start failed due **Compilation error**

cristianberroeta:
Use the run-script-os npm package to detect if the OS is Windows when running npm start. If so, it runs npm start:win32. Otherwise, it runs npm run start:default.

Also, instead of using the dotenv npm package to set the SASS_PATH variable in the .env file, now it is being set inline, in the npm scripts start:win32 and start:default.

Fix #210.


cristianberroeta:
hi @yuchenshi,

I see that a check failed. I suspect that could be caused because I removed the SASS_PATH variable from the .env file, and then only added it to the start npm scripts. However, that variable should probably be set for the build command, and maybe also the test commands. And if we want those commands (build and test) to work also on Windows, we should do the same trick that we did for the start script, by using run-script-os. It's not very clean but that should work.

cristianberroeta:
An alternative solution would be setting the SASS_PATH in the .env file, but to node_modules only; not to node_modules:src:
```bash
SASS_PATH=node_modules:src
```

By doing that, we would avoid having to deal with the different syntax that Windows requires for multiple paths. However, we would have to fix all the imports in the code that rely on having "src" in the SASS_PATH.

yuchenshi:
@cristianberroeta Would you mind updating the PR using `run-script-os` for the `build:web` command too? I don't think we need it for `test` or anything else.

cristianberroeta:
no problem, I can do that now 😄

cristianberroeta:
@yuchenshi, some minutes after I did the PR update, I received an email saying that the Node.js CI / unit (10.x) check had failed:
image

I was wondering what could have caused the failure, but then about 30 minutes later, I saw that the same check have passed.

What could have caused that?

yuchenshi (approved):

yuchenshi:
Node.js 10 tests are flaky for unknown reasons, but it's all green now after a rerun. I'll see if I can do some manual testing on OS X too

cristianberroeta:
ah, that's interesting

At least in my macOS High Sierra 10.13.6 the ```npm run build``` command seems to work correctly on branch cristianberroeta:fix-windows-issue-with-npm-start:
image

PR #643, opened 2021-09-07
Fix issue with the order of requests in the Firestore tab

### LINKED ISSUE #634: Firestore requests tab is unordered making it mostly unusable
The requests list at `/firestore/requests` is unordered. This makes it mostly unusable as it's very hard to find new requests as they happen. The page should be ordered by timestamp.

firebase-tools: 9.16.6
platform: MacOS arm64 (M1) 11.5.2
browser: Firefox 92.0b6

cristianberroeta:
By using the root time field, instead of the rulesContext.time field.

The former seems to have the timestamp at which the request was actually sent. On the other hand, the latter apparently keeps the timestamp at which some web socket communication was created. In the case of get-only-once firestore requests, that connection seems to last for 1 minute,

I had to add the time field to the FirestoreRulesEvaluation interface, as it wasn't there. And also, add the time field to the object returned by a testing util function.

Fix #634.

### DIFF
--- a/src/components/Firestore/Requests/RequestsCard/Table/TableRow.tsx
+++ b/src/components/Firestore/Requests/RequestsCard/Table/TableRow.tsx
@@ -41,7 +41,7 @@ function getTableRowRequestData(
 ): TableRowRequestData {
   const { rulesContext, outcome } = request;
   // time * 1000 converts timestamp units from seconds to millis
-  const requestDate = new Date(rulesContext.time);
+  const requestDate = new Date(request.time);
   return {
     requestTimeComplete: requestDate.toLocaleString(),
     requestTimeFormatted: requestDate.toLocaleTimeString('en-US', {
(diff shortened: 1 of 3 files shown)


src/components/Firestore/Requests/RequestsCard/Table/TableRow.tsx
```diff
@@ -41,7 +41,7 @@ function getTableRowRequestData(
 ): TableRowRequestData {
   const { rulesContext, outcome } = request;
   // time * 1000 converts timestamp units from seconds to millis
-  const requestDate = new Date(rulesContext.time);
+  const requestDate = new Date(request.time);
```
yuchenshi, line 44: Please use the destructuring assignment above on L42 and remove the outdated comment on L43. LGTM otherwise.

cristianberroeta: ah, good point @yuchenshi! I'll do that now

src/components/Firestore/testing/test_utils.ts
```diff
@@ -62,6 +62,7 @@ export function createFakeFirestoreRequestEvaluation(
     requestId: 'unique_id',
     granularAllowOutcomes: [],
     rules: SAMPLE_RULES,
+    time: new Date().toString(),
```
yuchenshi, line 65:
```suggestion
    time: new Date().toISOString(),
```

yuchenshi (approved):

PR #644, opened 2021-09-09
Remove `--only` flags from README instructions

jhuleatt:
The Storage tests failed when I ran `firebase emulators:exec --project sample --only firestore 'npm test'` because the Storage emulator wasn't started.

Tests seem to run fine without the `--only` flag, so I removed it from the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ We welcome any issues and Pull Requests to improve the Emulator UI. The followin
 To run the development server with test data:
 
 ```bash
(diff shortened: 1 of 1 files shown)


yuchenshi (changes requested):
The instructions were written before we had a `firebase.json`, so `--only` was the *only* way (pun intended) to start those emulators. I think your suggestion makes sense but we'd also need to add `database` into `firebase.json` for RTDB Emulator to be started.

PR #654, opened 2021-10-13
Add a passthrough mode toggle for Auth

jhuleatt:
Related to https://github.com/firebase/firebase-tools/pull/3624

### Screenshots

Zero state:

image

Enable passthrough dialog:

image

Passthrough mode enabled:

Screen Shot 2021-10-19 at 8 32 11 AM

Disable passthrough dialog:

Screen Shot 2021-10-19 at 8 27 05 AM

### DIFF
--- a/src/components/Auth/PassthroughModeCard/PassthroughModeCard.tsx
+++ b/src/components/Auth/PassthroughModeCard/PassthroughModeCard.tsx
@@ -1,0 +1,76 @@
+/**
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Button } from '@rmwc/button';
+import { Card } from '@rmwc/card';
+import { Typography } from '@rmwc/typography';
+import React, { useState } from 'react';
+import { connect } from 'react-redux';
+
+import { createStructuredSelector } from '../../../store';
+import { getUsageMode } from '../../../store/auth/selectors';
+import { UsageModes } from '../types';
+import styles from './PassthroughModeCard.module.scss';
+import PassthroughModeDialog from './PassthroughModeDialog';
+
+const ENABLED_HEADING = 'Passthrough mode enabled';
+const DISABLED_HEADING = 'Passthrough mode';
+
+const ENABLED_COPY =
+  'This mode is a form of custom auth that does not persist users in Firebase Auth emulator.';
+const DISABLED_COPY =
+  'This mode is a form of custom auth that does not persist users in Firebase Auth emulator.';
+
+export type PassthroughModeCardProps = PropsFromStore;
+export const PassthroughModeCard: React.FC<PassthroughModeCardProps> = ({
+  usageMode,
+}) => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <>
+      <Card className={styles.wrapper}>
+        <div>
+          <Typography use="headline6" tag="div" theme="textPrimaryOnBackground">
+            {usageMode === UsageModes.PASSTHROUGH ? ENABLED_HEADING : DISABLED_HEADING}
+          </Typography>
+          <Typography use="body2" theme="textPrimaryOnBackground">
+            {usageMode === UsageModes.PASSTHROUGH ? ENABLED_COPY : DISABLED_COPY}
+            <a
+              target="_blank"
+              rel="noopener noreferrer"
+              href="https://firebase.google.com/docs/auth"
+            >
+              Learn more
+            </a>
+          </Typography>
+        </div>
+        <div>
+          <Button outlined={true} onClick={() => setOpen(true)}>
+            {usageMode === UsageModes.PASSTHROUGH ? 'Disable' : 'Enable'}
+          </Button>
+        </div>
+      </Card>
+      {open && <PassthroughModeDialog onClose={() => setOpen(false)} />}
+    </>
+  );
+};
+export const mapStateToProps = createStructuredSelector({
+  usageMode: getUsageMode,
+});
+
+export type PropsFromStore = ReturnType<typeof mapStateToProps>;
+export default connect(mapStateToProps)(PassthroughModeCard);
--- a/src/components/Auth/PassthroughModeCard/PassthroughModeCard.tsx
+++ b/src/components/Auth/PassthroughModeCard/PassthroughModeCard.tsx
@@ -45,10 +45,14 @@ export const PassthroughModeCard: React.FC<PassthroughModeCardProps> = ({
       <Card className={styles.wrapper}>
         <div>
           <Typography use="headline6" tag="div" theme="textPrimaryOnBackground">
-            {usageMode === UsageModes.PASSTHROUGH ? ENABLED_HEADING : DISABLED_HEADING}
+            {usageMode === UsageModes.PASSTHROUGH
+              ? ENABLED_HEADING
+              : DISABLED_HEADING}
           </Typography>
           <Typography use="body2" theme="textPrimaryOnBackground">
-            {usageMode === UsageModes.PASSTHROUGH ? ENABLED_COPY : DISABLED_COPY}
+            {usageMode === UsageModes.PASSTHROUGH
+              ? ENABLED_COPY
+              : DISABLED_COPY}
             <a
               target="_blank"
               rel="noopener noreferrer"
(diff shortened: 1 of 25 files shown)


src/components/Auth/types.ts
```diff
@@ -62,9 +62,19 @@ export interface AuthUser extends AddAuthUserPayload {
...
+  // We shouldn't ever get this usage mode
+  USAGE_MODE_UNSPECIFIED = 'USAGE_MODE_UNSPECIFIED',
+}
+
 export interface AuthState {
   authUserDialogData?: RemoteResult<AuthUser | undefined>;
   users: RemoteResult<AuthUser[]>;
   filter: string;
   allowDuplicateEmails: boolean;
+  usageMode: UsageModes;
```
jhuleatt, line 79: Instead of an enum, we could make this a boolean instead (since `USAGE_MODE_UNSPECIFIED` shouldn't happen as far as I know). But that may make it trickier to handle more usage modes in the future, if that is a thing that is likely to happen.

src/store/auth/sagas.ts
```diff
@@ -166,16 +172,31 @@ export function* setAllowDuplicateEmails({
...
   const authApi: AuthApi = yield call(configureAuthSaga);
-  const config: boolean = yield call([authApi, 'getConfig']);
+  const config: boolean = yield call([authApi, 'getAllowDuplicateEmails']);
   yield put(setAllowDuplicateEmailsSuccess(config));
 }
 
+export function* getUsageMode() {
+  const authApi: AuthApi = yield call(configureAuthSaga);
+  const config: UsageModes = yield call([authApi, 'getUsageMode']);
+  console.log('saga - getUsageMode', config);
```
yuchenshi, line 196:
```suggestion
```

src/components/Auth/UsersCard/table/PassthroughMode.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const PassthroughMode: React.FC = () => {
+  return (
+    <Typography
+      use="body2"
+      className={styles.noResultsWrapper}
+      theme="textSecondaryOnBackground"
+    >
+      <span>
+        No users persist when passthrough mode is enabled.{' '}
```
yuchenshi, line 30: Any thoughts on how to make this sentence a bit more clear? Auth emulator doesn't "persist" users -- they are either in memory or "exported". @markarndt

src/store/auth/sagas.ts
```diff
@@ -166,16 +172,31 @@ export function* setAllowDuplicateEmails({
...
+}: ReturnType<typeof setUsageModeRequest>) {
+  const authApi: AuthApi = yield call(configureAuthSaga);
+  yield call([authApi, 'updateUsageMode'], payload);
+  yield put(setUsageModeSuccess(payload));
+}
+
 export function* getAllowDuplicateEmails() {
   const authApi: AuthApi = yield call(configureAuthSaga);
-  const config: boolean = yield call([authApi, 'getConfig']);
+  const config: boolean = yield call([authApi, 'getAllowDuplicateEmails']);
```
yuchenshi, line 189: nit: rename to `allowDuplicateEmails`

src/store/auth/sagas.ts
```diff
@@ -166,16 +172,31 @@ export function* setAllowDuplicateEmails({
...
   yield put(setAllowDuplicateEmailsSuccess(config));
 }
 
+export function* getUsageMode() {
+  const authApi: AuthApi = yield call(configureAuthSaga);
+  const config: UsageModes = yield call([authApi, 'getUsageMode']);
```
yuchenshi, line 195: nit: rename tp `usageMode`.

src/components/Auth/api.ts
```diff
@@ -87,22 +87,46 @@ export default class AuthApi extends RestApi {
...
+    return json.signIn?.allowDuplicateEmails;
   }
 
-  async getConfig(): Promise<AuthUser> {
-    const { signIn } = await (
-      await fetch(`${this.baseEmulatorUrl}/config`)
-    ).json();
+  async updateUsageMode(
+    usageMode: UsageModes
+  ): Promise<{ signIn: { allowDuplicateEmails: boolean } }> {
```
yuchenshi, line 102: typing here is wrong -- it should return `Promise<UsageModes>`
```suggestion
  ): Promise<UsageModes> {
```

src/components/Auth/api.ts
```diff
@@ -87,22 +87,46 @@ export default class AuthApi extends RestApi {
     return this.fetchUser(json.localId);
   }
 
-  async updateConfig(
+  async updateAllowDuplicateEmails(
```
yuchenshi, line 90: I'd actually suggest we should keep `updateConfig` as a single method here since the logic for both `allowDuplicateEmails` and `usageMode` is largely the same and it's one API endpoint after all. We can just make the parameter type `{ signIn?: { allowDuplicateEmails?: boolean }, usageMode?: UsageModes }` so callers can update whichever they like. The return type can be non-optional.

src/components/Auth/types.ts
jhuleatt: Updated to `UsageMode` and moved `USAGE_MODE_UNSPECIFIED` to the top.

src/components/Auth/UsersCard/table/PassthroughMode.tsx
jhuleatt: What about:

> This table is unavailable because passthrough mode is enabled. Manage users in your custom Auth system instead.

src/store/auth/sagas.ts
jhuleatt: done

src/store/auth/sagas.ts
jhuleatt: done

src/components/Auth/api.ts
jhuleatt: removed the `updateUsageMode` function per https://github.com/firebase/firebase-tools-ui/pull/654#discussion_r732312481

src/components/Auth/api.ts
jhuleatt: Makes sense to me, I've updated the API to only have `getConfig` and `updateConfig`.

src/components/Auth/UsersCard/table/PassthroughMode.tsx
jhuleatt: We should probably also update https://github.com/firebase/firebase-tools-ui/pull/654/files#diff-3f0b79d8c25faa664fe6f6a2f91a46da041f83608b6ca4a3e79fde9be88eabdcR33 if we change this language

src/components/Auth/api.ts
```diff
@@ -87,46 +91,20 @@ export default class AuthApi extends RestApi {
...
-
-  async updateUsageMode(
-    usageMode: UsageMode
-  ): Promise<{ signIn: { allowDuplicateEmails: boolean } }> {
-    const { json } = await this.patchRequest(`${this.baseEmulatorUrl}/config`, {
-      usageMode,
-    });
-
-    return json.usageMode;
+  async updateConfig(newConfig: {
```
yuchenshi, line 94: nit: just use `newConfig: Partial<EmulatorV1ProjectsConfig>`

yuchenshi (approved):

src/components/Auth/PassthroughModeCard/PassthroughModeCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+            {usageMode === UsageMode.PASSTHROUGH
+              ? ENABLED_HEADING
+              : DISABLED_HEADING}
+          </Typography>
+          <Typography use="body2" theme="textPrimaryOnBackground">
+            {usageMode === UsageMode.PASSTHROUGH ? ENABLED_COPY : DISABLED_COPY}{' '}
+            <a
+              target="_blank"
+              rel="noopener noreferrer"
+              href="https://firebase.google.com/docs/auth"
```
yuchenshi, line 57: Note for later: this need to link to a more relevant doc page (once we publish it)

src/components/Auth/api.ts
jhuleatt: Updated, thanks! TIL about `Partial`

src/components/Auth/UsersCard/table/PassthroughMode.tsx
yuchenshi: I'm fine merging as-is (but please use a different branch like `auth-new-features` instead). We can run a tech writer review round later. These strings are easy to change anyway.

jhuleatt: will do

jhuleatt: new base branch: https://github.com/firebase/firebase-tools-ui/tree/auth-new-features

src/components/Auth/PassthroughModeCard/PassthroughModeCard.tsx
jhuleatt: added a TODO so we're less likely to forget

src/components/Auth/UsersCard/table/PassthroughMode.tsx
markarndt: How about "...does not save user authentication credentials when passthrough is enabled...."?

yuchenshi: @markarndt Love that! I'd just say "does not save user information" though and that should cover what we're doing here. I'll draft a PR for the copy change.

yuchenshi: https://github.com/firebase/firebase-tools-ui/pull/659

PR #659, opened 2021-11-08
Update passthough mode text copies.

yuchenshi:


src/components/Auth/PassthroughModeCard/PassthroughModeCard.tsx
```diff
@@ -30,9 +30,12 @@ const ENABLED_HEADING = 'Passthrough mode enabled';
...
 
 export const ENABLED_COPY =
-  'This mode is a form of custom auth that does not persist users in Firebase Auth emulator.';
+  'All sign-in methods are disabled except Custom Token. User information is not saved in the emulator.';
+
+// Consider updating the copy with the canonical short description of
+// passthrough mode, once published in Firebase Auth docs.
 export const DISABLED_COPY =
-  'This mode is a form of custom auth that does not persist users in Firebase Auth backend.';
+  'This mode is a form of custom auth that does not save user information.';
```
jhuleatt, line 38: I'm a little worried "does not save user information" might get confused with GDPR compliance or similar. Can we highlight the tradeoff a bit more? Something like "Use this mode if you'd rather store users in your own Auth system" (if that's accurate)?

markarndt: DISABLED_COPY: is this UI copy? How much verbiage do we have room for? 

What about "...does not save user information in the Firebase Auth backend" to make the literal truth of where thngs are or not stored clear.

Now, beyond that, something more explanatory, do we need something like "...does not save user information in the Firebase Auth backend, though it is saved in your custom auth solution."  ?

markarndt: Also, wrt to GDPR, what's in the Firebase ID token that users may feel misled about if we claim it it or isn't stored?

yuchenshi: @markarndt There are three sentences in this PR. The DISABLED_COPY and ENABLED_COPY will be shown next to a toggle button similar to this:

Screen Shot 2021-11-09 at 3 54 07 PM

The exact screen space depends on window width, but we can definitely fit a longer sentence. Preferably without wrapping.

The other sentence is for the zero state of the users table, which I put "The emulator does not save user information when passthrough mode is enabled" there. This should be short due to the space. It does not need to be too elaborate though since developers will only see this after they enable passthrough (i.e. already seen the copies above). Feel free to take a look at the [mocks](https://www.figma.com/file/nYaYu2XflZqwrd9xyCCoiD/Passthrough-mode?node-id=1%3A2651) on how these three lines will be displayed.

yuchenshi: All copies updated based on internal discussions.

jhuleatt (approved):

markarndt (commented):
Approved. (I don't see the Comment, Approve, Request changes options when reviewing PRs in this repo)

PR #683, opened 2022-01-27
Bugbash fixes

jhuleatt:
- [x] delete users before enabling Passthrough mode
- [x] allow save on user edit
- [x] disable save until some authentication method is provided

### DIFF
--- a/src/store/auth/sagas.ts
+++ b/src/store/auth/sagas.ts
@@ -20,6 +20,7 @@ import {
   getContext,
   put,
   setContext,
+  take,
   takeLatest,
 } from 'redux-saga/effects';
 import { ActionType, getType } from 'typesafe-actions';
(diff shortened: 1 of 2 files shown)


src/store/auth/sagas.ts
```diff
@@ -184,6 +184,12 @@ export function* setUsageMode({
   payload,
 }: ReturnType<typeof setUsageModeRequest>) {
   const authApi: AuthApi = yield call(configureAuthSaga);
+
+  // passthrough mode can't be enabled until all users are deleted
+  if (payload === 'PASSTHROUGH') {
+    yield put(nukeUsersRequest());
```
yuchenshi, line 190:
```suggestion
    yield put(nukeUsersRequest());
    yield take(getType(nukeUsersSuccess));
```

Gotta wait until users are actually deleted

yuchenshi (approved):

PR #696, opened 2022-02-07
Fix mfa validation

jhuleatt:
screen recording of updated validation logic: go/firebase-tools-ui-mfa-validation

internal tracking bug: b/216660333

### DIFF
--- a/src/components/Auth/UserFormDialog/UserForm.tsx
+++ b/src/components/Auth/UserFormDialog/UserForm.tsx
@@ -62,26 +62,34 @@ function convertToFormUser(user?: AuthUser): AuthFormUser | undefined {
 }
 
 function convertFromFormUser(formUser: AuthFormUser): AddAuthUserPayload {
+  // remove any empty MFA phoneInfo fields
+  const filteredMfaPhoneInfo =
+    formUser.mfaPhoneInfo &&
+    formUser.mfaPhoneInfo.filter(
+      (mfaPhoneInfo) => mfaPhoneInfo.phoneInfo !== ''
+    );
+
   return {
     ...formUser,
     emailVerified: formUser.emailVerified.length > 0 ? true : false,
-    // match mfaPhoneInfo array members to an existing enrollment, or create a new enrollment
-    mfaInfo: formUser.mfaPhoneInfo
-      ? formUser.mfaPhoneInfo.map((mfaPhoneInfo) => {
-          const existingEnrollment = formUser.mfaInfo?.find(
-            (mfaEnrollment) =>
-              mfaEnrollment.phoneInfo === mfaPhoneInfo.phoneInfo
-          );
-          return (
-            existingEnrollment || {
-              ...mfaPhoneInfo,
-              enrolledAt: new Date().toISOString(),
-              mfaEnrollmentId:
-                'AUTH-EMULATOR-UI:' + Math.random().toString(36).substring(5),
-            }
-          );
-        })
-      : undefined,
+    mfaInfo:
+      filteredMfaPhoneInfo.length > 0
+        ? // match mfaPhoneInfo array members to an existing enrollment, or create a new enrollment
+          filteredMfaPhoneInfo.map((mfaPhoneInfo) => {
+            const existingEnrollment = formUser.mfaInfo?.find(
+              (mfaEnrollment) =>
+                mfaEnrollment.phoneInfo === mfaPhoneInfo.phoneInfo
+            );
+            return (
+              existingEnrollment || {
+                ...mfaPhoneInfo,
+                enrolledAt: new Date().toISOString(),
+                mfaEnrollmentId:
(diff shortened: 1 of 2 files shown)


lisajian (commented):
Hey Jeff! Could you grant access to the go/ link in the PR description? I'm hitting a message that says I'm not unauthorized to view the recording

jhuleatt:
oops, think I fixed the permissions now

lisajian (approved):
Overall LGTM! Yuchen has more context on MFA since he worked on adding the functionality to the Emulator Suite, so you may want to wait to get an LGTM from him too

src/components/Auth/UserFormDialog/controls/MultiFactorAuth.tsx
```diff
@@ -148,9 +144,10 @@ export const MultiFactor: React.FC<
                     placeholder="Enter phone number"
                     type="tel"
                     error={getPhoneErrorText()}
+                    disabled={mfaEnabled === false}
```
yuchenshi, line 147:
```suggestion
                    disabled={!mfaEnabled}
```

`mfaEnabled` may be `undefined` (see above)

src/components/Auth/UserFormDialog/controls/MultiFactorAuth.tsx
```diff
@@ -148,9 +144,10 @@ export const MultiFactor: React.FC<
...
                     inputRef={register({
                       pattern: PHONE_REGEX,
-                      required: true,
+                      required: mfaEnabled === true,
```
yuchenshi, line 150:
```suggestion
                      required: !!mfaEnabled,
```

src/components/Auth/UserFormDialog/UserForm.tsx
```diff
@@ -65,23 +65,23 @@ function convertFromFormUser(formUser: AuthFormUser): AddAuthUserPayload {
...
-            existingEnrollment || {
-              ...mfaPhoneInfo,
-              enrolledAt: new Date().toISOString(),
-              mfaEnrollmentId:
-                'AUTH-EMULATOR-UI:' + Math.random().toString(36).substring(5),
-            }
-          );
-        })
-      : undefined,
+    mfaInfo:
```
yuchenshi, line 68: Please extract this into a variable. Please also remove the ternary while we're at it.

```js
let mfaInfo: AddAuthUserPayload["mfaInfo"];
if (formUser.mfaEnabled[0] === 'on') {
  mfaInfo = ...
}
```

src/components/Auth/UserFormDialog/UserForm.tsx
```diff
@@ -65,23 +65,23 @@ function convertFromFormUser(formUser: AuthFormUser): AddAuthUserPayload {
   return {
     ...formUser,
```
yuchenshi, line 66: I wonder if we should remove extra fields like `mfaPhoneInfo` from the payload as well? Those fields are only used in the form and won't make sense when passed into the API.

yuchenshi:
I've got one more fix for you :) #699

src/components/Auth/UserFormDialog/UserForm.tsx
jhuleatt: I believe they're gone already, since this function returns an `AddAuthUserPayload`: https://github.com/firebase/firebase-tools-ui/blob/211bd5bbcf85ef2f8c47a6e4401ae087aea16f6c/src/components/Auth/types.ts#L41-L51

Which contains these:

https://github.com/firebase/firebase-tools-ui/blob/211bd5bbcf85ef2f8c47a6e4401ae087aea16f6c/src/components/Auth/types.ts#L29-L35

Or am I missing it somewhere else?

src/components/Auth/UserFormDialog/controls/MultiFactorAuth.tsx
jhuleatt: done

src/components/Auth/UserFormDialog/controls/MultiFactorAuth.tsx
jhuleatt: done

src/components/Auth/UserFormDialog/UserForm.tsx
yuchenshi: `formUser` is of type `AuthFormUser`, which contains three extra fields: `mfaPhoneInfo`, `mfaEnabled` and `emailVerified`. The code here copies over them (by splating `formUser`) and then overrides `emailVerified`, but I believe `mfaEnabled` and `mfaPhoneInfo` are still included in the resulting object.

Note that `mfaPhoneInfo`, unlike `mfaInfo`, isn't part of the API contract. (Also note that TypeScript won't catch these extra fields since this object is technically compatible with `AddAuthUserPayload` since it has a super set of fields. TS only complains about object literals having extra fields when splats are not used)

yuchenshi (approved):

PR #713, opened 2022-02-22
Clear MFA validation errors when MFA is disabled

jhuleatt:


src/components/Auth/api.ts
```diff
@@ -54,8 +55,34 @@ export default class AuthApi extends RestApi {
...
+    const tenants = await this.fetchTenants();
+    const tenantDeletes = tenants.map(({ tenantId }) => {
+      return this.deleteRequest(
+        `${this.baseEmulatorUrl}/tenants/${tenantId}/accounts`
+      );
+    });
+    deletePromises.concat(tenantDeletes);
+
+    await Promise.all(deletePromises);
+    return [];
```
yuchenshi, line 74: nit: don't return this and just change function signature to return `Promise<void>`

yuchenshi (approved):

PR #717, opened 2022-02-24
Add basic API to fetch list of installed extensions

tjlav5:


src/components/Extensions/models.ts
```diff
@@ -14,29 +14,49 @@
  * limitations under the License.
  */
 
-export type ExtensionName = `publishers/${string}/extensions/${string}`;
+// TODO(tjlav5): re-enable name-types when new babel version lands
```
NothingEverHappens, line 17: is there a bug or an issue for that?

NothingEverHappens (approved):

src/components/Extensions/models.ts
tjlav5: I couldn't find an exact bug, but I found this: https://github.com/facebook/jest/issues/11022. And it seems like CRA is really the issue here since it include the dep on babel. I think once we upgrade CRA to 5, we should be OK

PR #719, opened 2022-02-25
Add an empty extension route

NothingEverHappens:


tjlav5:
Is this safe to drop?

PR #727, opened 2022-03-15
Fix storage emulator tests

tonyjhuang:
Since https://github.com/firebase/firebase-tools/pull/3647 was merged, the list endpoint now exhibits different (more correct) behavior than what the integration tests in this repo was testing against. This PR fixes the assertions in the integration tests to match the new live version of the storage
 emulator.

Note: https://github.com/firebase/firebase-tools/pull/3647 introduces a new visual bug in the emulator wherein zero-byte files that denote empty folders are now shown in the file browser. 

Screen Shot 2022-03-16 at 9 20 21 PM

The behavior of the storage emulator is correct in this case, but the view logic of the UI needs to filter out these files before rendering. Files https://github.com/firebase/firebase-tools-ui/issues/728 to track.

### DIFF
--- a/src/components/Storage/Card/Table/Header/StorageHeader/CreateFolder/CreateFolder.test.tsx
+++ b/src/components/Storage/Card/Table/Header/StorageHeader/CreateFolder/CreateFolder.test.tsx
@@ -25,7 +25,7 @@ import { renderWithStorage } from '../../../../../testing/renderWithStorage';
 import { CreateFolder } from './CreateFolder';
 
 describe('CreateFolder', () => {
-  it.skip('opens folder dialog and then creates folder', async () => {
+  it('opens folder dialog and then creates folder', async () => {
     const {
       getByLabelText,
       getByText,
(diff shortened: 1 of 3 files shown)


src/components/Storage/Card/Table/Header/StorageHeader/CreateFolder/CreateFolder.test.tsx
```diff
@@ -25,7 +25,7 @@ import { renderWithStorage } from '../../../../../testing/renderWithStorage';
 import { CreateFolder } from './CreateFolder';
 
 describe('CreateFolder', () => {
-  it('opens folder dialog and then creates folder', async () => {
+  it.skip('opens folder dialog and then creates folder', async () => {
```
yuchenshi, line 28: please add a comment

yuchenshi (approved):
Approved, but other tests still seem flaky on CI

src/components/Storage/Card/Table/Header/StorageHeader/CreateFolder/CreateFolder.test.tsx
tonyjhuang: No longer skipping this test.

abhis3 (approved):
LGTM

yuchenshi (approved):

PR #769, opened 2022-05-17
Fix #601. Cannot enter timestamp field on document in Firestore

### LINKED ISSUE #601: Cannot enter timestamp field on document in Firestore
Firebase tools version: 9.12.1
Safari 14.1.1
macOS Big Sur

When creating a timestamp field entry on a document using the Firestore Emulator UI, the autocorrect logic provided by Safari prevents saving the new field. No date I enter is accepted by the autocorrect logic, preventing me from pressing save.

Screenshot 2021-06-28 at 9 37 35 am

gianlucatomasino:
The datetime-local input type has a default step of 60 (1 minute). 
Since the defaultValue is formatted also with seconds this create the issue.
Set step to 1 resolve issue #601.


yuchenshi (approved):
Wow thanks!

yuchenshi:
Thanks again for the contribution. As you've probably seen, we've had some issues with the CI but we'll release again with your fix once we get the build to a more stable situation again.

PR #774, opened 2022-05-23
Tl/form hooks v7

tjlav5:
Looks like some more react 18 issues here where react-hook-form is explicitly limited to react 16/17 on older versions

PR into the node branch as this should get the build green all in one-go


yuchenshi (approved):
Great work! There are still some flaky unit tests though, maybe we can address later

PR #788, opened 2022-05-26
Immer v9

tjlav5:
Updates to immer v9 and some changes to address types not working with 'typesafe-actions'-lib

### DIFF
--- a/src/store/auth/reducer.tsx
+++ b/src/store/auth/reducer.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import produce, { produceWithPatches } from 'immer';
+import produce from 'immer';
 import { Action, createReducer } from 'typesafe-actions';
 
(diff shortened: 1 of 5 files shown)


src/store/auth/reducer.tsx
```diff
@@ -28,106 +28,96 @@ const INIT_STATE = {
...
-  .handleAction(
-    authActions.setAuthUserDialogError,
-    produce((draft, { payload: error }) => {
+    });
+  })
+  .handleAction(authActions.setAuthUserDialogError, (state, { payload }) => {
+    return produce(state, (draft) => {
       if (draft.authUserDialogData && draft.authUserDialogData.result) {
-        draft.authUserDialogData.result.error = error;
+        (draft.authUserDialogData.result as any).error = payload;
```
NothingEverHappens, line 99: why do you need any here?
Is this because of the readonly error?
https://immerjs.github.io/immer/typescript#cast-utilities

PR #791, opened 2022-05-26
Set legacy ipv4first for dns resolution

tjlav5:
Bumps heap size for our tests where they were skirting a 2gb limit and crashing.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -63,7 +63,7 @@ jobs:
         # forceExit: something is not cleaning up - likely firebase-app that needs to be deleted, investigate
         run: |
           npm install -g firebase-tools@latest
-          firebase emulators:exec --only database,firestore,storage --project sample 'npm run test:coverage -- --forceExit --logHeapUsage'
+          firebase emulators:exec --only database,firestore,storage --project sample 'npm run test:coverage -- --forceExit'
         env:
           CI: true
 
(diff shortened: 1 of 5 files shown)


.github/workflows/nodejs.yml
```diff
@@ -60,7 +63,7 @@ jobs:
         # forceExit: something is not cleaning up - likely firebase-app that needs to be deleted, investigate
         run: |
           npm install -g firebase-tools@latest
-          firebase emulators:exec --only database,firestore,storage --project sample 'npm run test:coverage -- --forceExit'
+          firebase emulators:exec --only database,firestore,storage --project sample 'npm run test:coverage -- --forceExit --logHeapUsage'
         env:
```
yuchenshi, line 67: Is this still needed now we know the memory usage?

package.json
```diff
@@ -81,7 +80,7 @@
     "build:server": "webpack --target=node ./server.js -o server.bundle.js",
     "test": "react-scripts test --env=./custom-jest-environment.js",
     "test:debug": "react-scripts --inspect-brk test --env=./custom-jest-environment.js --runInBand",
-    "test:coverage": "react-scripts test --coverage --watchAll=false --env=./custom-jest-environment.js",
+    "test:coverage": "react-scripts --expose-gc test --coverage --watchAll=false --env=./custom-jest-environment.js",
```
yuchenshi, line 83: Ditto

yuchenshi (approved):

.github/workflows/nodejs.yml
tjlav5: Hmm technically no - but it might help keep our tests in check, if we notice the size spiking? Granted no one will ever notice it so maybe not worth keeping. I can go either way.

Argument for keeping: future debugging is quicker
Argument for dropping: not really needed for the coverage test

wdyt?

PR #852, opened 2022-09-02
Add back type checking and fix errors.

yuchenshi:
Warnings are split into a follow-up PR #853 (which touches a lot of files).


.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 /**
```
yuchenshi, line 1: This is not a rename, just deleting one file and adding another file. git got this wrong here

PR #667, opened 2021-12-03
fix: replace hardcoded base url protocol to use client-side protocol

BobbieGoede:
Fixes: #666

PR #858, opened 2022-10-05
Support emulators on HTTPS. Fixes #666.

### LINKED ISSUE #666: Does not support emulator urls using HTTPS
For local development I'm trying to run firebase emulators in a docker container, this works fine, but the dashboard does not work since the base urls are hardcoded to be insecure. I'm not sure if this is intended to prevent running emulators behind a reverse proxy with ssl.

Because the dashboard will run on a secure url, insecure api calls will fail like so:
```
api.ts:102 Mixed Content: The page at 'https://domain.test:4050/auth' was loaded over HTTPS, but requested an insecure resource 'http://domain.test:9099/emulator/v1/projects/project-id/config'. This request has been blocked; the content must be served over HTTPS.
```

Base url protocol is hardcoded as seen here:
https://github.com/firebase/firebase-tools-ui/blob/3ef507ed499c4f1ad4c6a60926b12f2f7534c54b/src/components/Auth/api.ts#L40

yuchenshi:
This replaces #667.


christhompsongoogle (approved):

davgothic:
Does anyone know where to track the releases of this project? I believe at the minute version 1.9.0 is installed when issuing `firebase emulators init`.

But I'd like to know when this feature will land in the install as I'm keen to try it out and there doesn't appear to be any public release tracker.

yuchenshi:
@davgothic The releases are bundled with the Firebase CLI so we don't have a dedicated tracker. You can instead navigate to https://github.com/firebase/firebase-tools and watch the repo there -- you can customize what to watch, e.g. only releases.

In case you wonder, there's no fixed release schedule either but we're cutting a release right now. This is your chance to test if your tracking method is working :-)

davgothic:
> @davgothic The releases are bundled with the Firebase CLI so we don't have a dedicated tracker. You can instead navigate to https://github.com/firebase/firebase-tools and watch the repo there -- you can customize what to watch, e.g. only releases.
> 
> In case you wonder, there's no fixed release schedule either but we're cutting a release right now. This is your chance to test if your tracking method is working :-)

Brilliant, thanks for the reply!

yuchenshi:
Released in CLI v11.14.0. It's not perfect, especially not all emulators work over HTTPS, but we think it's a step in the right direction. Let us know if you run into issues and PRs are welcomed!

PR #865, opened 2022-10-10
Fix websocket host logic.

yuchenshi:
The previous code incorrectly issues warnings when the hosts match and did not apply on the fix. The new logic will effectively ignore the `webSocketHost` field -- a different configuration isn't supported by the CLI anyway.


christhompsongoogle (approved):
Nit: please update the description to describe the old/new behavior

yuchenshi:
> Nit: please update the description to describe the old/new behavior

Updated, thanks for the reminder

PR #881, opened 2022-12-07
fix(a11y) Show focus/hover shadow/ipple effect on BreadCrumb Buttons

TheIronDev:
This commit adds back the ripple effect for buttons within the Breadcrumb area.

This fixes the issue where the "Import JSON" button was not accessible. In truth, the button was always accessible, but without the ripple/shadow the user doesn't know they are focused on the button.

FIXED=259452701, b/259452701
b/259452701


christhompsongoogle (approved):
Good observation, it is focusable just no animation

PR #883, opened 2022-12-07
fix(a11y) Add "enter" ability for logs page

TheIronDev:
This commit updates the log query textarea into an input and adds the "Enter" functionality."

The "input" part is almost unncessary, since we could have accomplished the behavior by just adding the "onKeyUp", but then we are conveying a wrong message to screen reader users based on the element we are using.

FIXED=259450050, b/259450050
b/259450050


yuchenshi (approved):
LGTM after reverting to input

PR #893, opened 2022-12-13
Use address instead of host when determining the URL to use.

aaron-coalesce:
Querying the hub for the emulator config by requesting`127.0.0.1:4400/emulators` gives the following:

```json
{
  "hub": {
    "listen": [
      {
        "address": "0.0.0.0",
        "family": "IPv4",
        "port": 4400
      }
    ],
    "name": "hub",
    "host": "127.0.0.1",
    "port": 4400
  },
  "ui": {
    "listen": [
      {
        "address": "0.0.0.0",
        "family": "IPv4",
        "port": 4000
      }
    ],
    "name": "ui",
    "host": "127.0.0.1",
    "port": 4000,
    "pid": 66
  }
}
```

The `listen.address` is `0.0.0.0`, so I would expect the UI logic to default to `window.location.hostname` but instead the links I see in the UI are `127.0.0.1`. This appears to be because the code is looking at `listen.host` instead of `listen.address`. Changing references to `listen.host` to `listen.address` fixed the issue, and the UI links used the window's location instead of `127.0.0.1`.

### DIFF
--- a/src/components/common/EmulatorConfigProvider.tsx
+++ b/src/components/common/EmulatorConfigProvider.tsx
@@ -217,10 +217,10 @@ async function configFetcher(url: string): Promise<Config> {
     // remotely and connect from another device. It does not work for some setups
     // e.g. when Emulator UI is proxied but not the individual emulators.
     for (const listen of (value.listen ?? []) as {
-      host: string;
+      address: string;
       port: number;
     }[]) {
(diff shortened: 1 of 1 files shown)


aaron-coalesce:
Looks like this got incorporated without me signing the agreement, so just closing the PR.

PR #915, opened 2023-01-27
fix: updating rc template should reflect on emulator

HYACCCINT:


src/components/RemoteConfig/index.tsx
```diff
@@ -41,21 +41,34 @@ function RemoteConfig() {
...
 
   const [searchText, setSearchText] = useState('');
+  const [edited, setEdited] = useState(false);
   const [paramBeingEdited, editParam] = useState<string | undefined>(undefined);
   const [editTemplate, setEditTemplate] = useState(
     JSON.parse(JSON.stringify(template))
   );
 
-  const editing = isEqual(template, editTemplate);
+  const editing = isEqual(template, editTemplate) && !edited;
```
jhuleatt, line 50: This seems backwards to me - should it be:

```suggestion
  const editing = !isEqual(template, editTemplate) && edited;
```

src/components/RemoteConfig/index.tsx
```diff
@@ -41,21 +41,34 @@ function RemoteConfig() {
...
+
+  useEffect(() => {
```
jhuleatt, line 52: please add a comment to make it clear what this is doing (reset the edit template when a new template comes in from the server?)

jhuleatt (approved):
Looks great! A couple small comments to address before merging

src/components/RemoteConfig/index.tsx
jhuleatt: should this be a `useMemo` instead of a `useEffect`?

PR #928, opened 2023-03-17
Update each emulator page with a title and heading.

christhompsongoogle:


yuchenshi (approved):
+1 to this change, but have we considered making some of the titles screen-reader-only? It may help us regain a bit of vertical space (especially on the Firestore viewer page)

PR #937, opened 2023-04-05
Adjust labels on navigation elements for users using a screen reader.

christhompsongoogle:
Internal bugs: b/275114898 and b/276523147.


yuchenshi (approved):

src/components/Auth/UserFormDialog/controls/CustomAttributes.tsx
```diff
@@ -81,6 +81,7 @@ export const CustomAttributes: React.FC<
           target="_blank"
           rel="noopener noreferrer"
           href="https://firebase.google.com/docs/auth/admin/custom-claims"
+          aria-label="Learn more about custom key value attributes."
```
yuchenshi, line 84: nit: Consider using `title` here since the text is helpful for most user agents (such as mouse hover tooltip), not just screen readers

PR #946, opened 2023-05-10
Update PanelHeader.scss remove overflow property

maddoxmarius:
Unnecessary space added due to the `overflow` property

image


maddoxmarius:
This is already fixed

PR #949, opened 2023-05-15
Move react-scripts out of prod dependencies and into devDependencies to mitigate a warning in prod.

christhompsongoogle:
Example:
https://github.com/firebase/firebase-tools-ui/security/dependabot/48


yuchenshi (approved):
I think this is fine given that the Emulator UI is distributed as a bundle which doesn't use package.json

PR #951, opened 2023-05-16
Fixed a a11y issue with the extensions page decorative icons being read.

christhompsongoogle:
The icon itself is a built-in of the rmwc Button and so I made the button visible instead of visible-on-hover and removed the icon.

Internal bug: b/277826813.

### DIFF
--- a/src/components/Extensions/List/ExtensionsTable/ExtensionTableRow.tsx
+++ b/src/components/Extensions/List/ExtensionsTable/ExtensionTableRow.tsx
@@ -73,7 +73,7 @@ export const ExtensionsTableRow: React.FC<
           }}
           className={styles.manageButton}
           label={<div className={styles.manageButtonText}>Manage</div>}
-          aria-label={'Manage, ' + extension.displayName}
+          aria-label={'Manage ' + extension.displayName}
         />
       </DataTableCell>
     </DataTableRow>
--- a/src/components/Extensions/List/ExtensionsTable/ExtensionTableRow.tsx
+++ b/src/components/Extensions/List/ExtensionsTable/ExtensionTableRow.tsx
@@ -73,7 +73,7 @@ export const ExtensionsTableRow: React.FC<
           }}
           className={styles.manageButton}
           label={<div className={styles.manageButtonText}>Manage</div>}
-          trailingIcon="arrow_forward"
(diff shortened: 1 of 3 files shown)


src/components/Extensions/List/ExtensionsTable/ExtensionTableRow.tsx
```diff
@@ -73,7 +73,7 @@ export const ExtensionsTableRow: React.FC<
           }}
           className={styles.manageButton}
           label={<div className={styles.manageButtonText}>Manage</div>}
-          trailingIcon="arrow_forward"
+          aria-label={'Manage, ' + extension.displayName}
```
yuchenshi, line 76:
```suggestion
          aria-label={'Manage ' + extension.displayName}
```

src/components/common/links/ExternalLink.tsx
```diff
@@ -33,7 +33,7 @@ export function ExternalLink(
       rel="noopener noreferrer"
     >
       <span className={styles.content}>{props.children}</span>
-      <Icon icon="open_in_new" className={styles.icon} />
+      <Icon aria-hidden="true" icon="open_in_new" className={styles.icon} />
```
yuchenshi, line 36:
```suggestion
      <Icon alt="" icon="open_in_new" className={styles.icon} />
```

go/GAR-web#overview-of-criterion-11 Best practices 3: You can hide decorative or redundant images from assistive technology by setting the alternative text to an empty string (alt="")

I'm not 100% confident that this will work on a font icon though. Please test and inspect the accessibility tree -- I'm curious about the findings

yuchenshi (approved):

src/components/common/links/ExternalLink.tsx
yuchenshi: If it generates HTML as a <span> or similar then `alt` won't work. Feel free to close this comment in that case

christhompsongoogle: Testing this after a long break, the generated HTML is below. Adding alt doesn't prevent the reader from reading the icon, likely because there's some inner text:


`<i aria-hidden="true" class="rmwc-icon rmwc-icon--ligature material-icons _icon_1cowi_23">open_in_new</i>`

PR #952, opened 2023-05-16
Fixed an issue where some headings weren't using an explicit header tag.

christhompsongoogle:
Necessary for screen reader navigation.

Internal bug: b/277815251


src/index.tsx
```diff
@@ -60,6 +60,10 @@ root.render(
   <RMWCProvider
     // Globally disable ripples
     ripple={false}
+    typography={{
+      /** Make headline5 <h5>  */
```
yuchenshi, line 64: How does this comment match the code below?

yuchenshi (approved):

src/index.tsx
christhompsongoogle: Fixed!

PR #959, opened 2023-06-27
A11y Bug Fix: Screenreader Announcement on `Clear All Data`

ABradham:
Fixed an accessibility bug where screen readers wouldn't report that all users had been cleared from the table after the `Clear All Data` button on the `Authentication` screen was pressed.

Fixed by adding a [visually hidden](https://www.a11yproject.com/posts/how-to-hide-content/) [aria alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) that notifies screen readers without changing the UI.

This change also required the addition of a new field to the `AuthState` redux store.

### DIFF
--- a/src/components/Auth/UsersCard/table/UsersCleared.test.tsx
+++ b/src/components/Auth/UsersCard/table/UsersCleared.test.tsx
@@ -17,11 +17,11 @@
 import { render } from '@testing-library/react';
 import React from 'react';
 
-import UsersCleared from './UsersCleared';
+import { UsersCleared } from './UsersCleared';
 
 describe('UsersCleared', () => {
   it('renders header row when all users have been cleared from the table', () => {
-    const { getByRole } = render(<UsersCleared />);
+    const { getByRole } = render(<UsersCleared alertText="" />);
     expect(getByRole('alert')).not.toBeNull();
   });
 });
--- a/src/components/Auth/UsersCard/table/UsersCleared.test.tsx
+++ b/src/components/Auth/UsersCard/table/UsersCleared.test.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021 Google LLC
+ * Copyright 2023 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,7 @@
 import { render } from '@testing-library/react';
 import React from 'react';
 
-import { UsersCleared } from './UsersCleared';
+import UsersCleared from './UsersCleared';
 
 describe('UsersCleared', () => {
   it('renders header row when all users have been cleared from the table', () => {
--- a/src/components/Auth/UsersCard/table/UsersCleared.test.tsx
+++ b/src/components/Auth/UsersCard/table/UsersCleared.test.tsx
@@ -22,6 +22,6 @@ import { UsersCleared } from './UsersCleared';
 describe('UsersCleared', () => {
   it('renders header row when all users have been cleared from the table', () => {
     const { getByText } = render(<UsersCleared />);
-    expect(getByText(/All users cleared/)).not.toBeNull();
+    expect(getByText('')).not.toBeNull();
   });
 });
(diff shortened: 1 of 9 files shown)


src/components/Auth/UsersCard/table/UsersTable.tsx
```diff
@@ -169,6 +172,7 @@ export const UsersTable: React.FC<React.PropsWithChildren<UsersTableProps>> = ({
           </DataTableBody>
         </DataTableContent>
       </DataTable>
+      {shouldShowNukedUsers && <UsersCleared />}
```
christhompsongoogle, line 175: nit: let's remove the nuked terminology for "cleared" or "clearedAll" for clarity

ABradham: For sure! I also found the term "nuked" kind of gauche, but kept it to follow with the action definitions in `src/store/auth/actions.tsx`. I'll change the names to `cleared`!

ABradham:
Added `alertText` field to redux store and refactored `UsersCleared` component to fetch props directly from store (no passing down props from parent component or conditional rendering from parent component). This pattern seems extensible to other alerts; It might make sense to rename the component from  `UsersCleared` to `UserAlerts` so that other actions that should notify screen readers (creating users, fetching data from the store, starting the emulator, etc.) can take advantage of this component / pattern as well.

src/components/Auth/UsersCard/table/UsersCleared.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render } from '@testing-library/react';
+import React from 'react';
```
christhompsongoogle, line 18: This bit shouldn't be necessary anymore, after react 17 or so

src/components/Auth/UsersCard/table/UsersCleared.tsx
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright 2021 Google LLC
```
christhompsongoogle, line 2: nit: 2023

christhompsongoogle: Please update the other headers x3

src/components/Auth/UsersCard/table/UsersCleared.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { createStructuredSelector } from '../../../../store';
+import { getAlertText } from '../../../../store/auth/selectors';
+import styles from './UsersCleared.module.scss';
+
+export type UsersClearedProps = PropsFromStore;
+
+export const UsersCleared: React.FC<
+  React.PropsWithChildren<UsersClearedProps>
+> = ({ alertText }) => {
```
christhompsongoogle, line 27: format

src/store/auth/reducer.tsx
```diff
@@ -35,6 +36,7 @@ export const authReducer = createReducer<AuthState, Action>(INIT_STATE)
         ...users,
         payload.user,
       ]);
+      draft.alertText = ''; // Maybe have an alert for "User Created"?
```
christhompsongoogle, line 39: Best to use TODOs:

// TODO(abradham): Consider adding an alert for "User Created"

christhompsongoogle (approved):

src/components/Auth/UsersCard/table/UsersCleared.test.tsx
ABradham: I get this error on the `<UsersCleared />` component when I remove the import: `'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.`
 
I'm not entirely sure why importing React is necessary here, but it seems to break the build when I remove it.

src/components/Auth/UsersCard/table/UsersCleared.tsx
ABradham: This seems to be the way that the `Prettier` formatter renders this unwieldy component definition; This is the output after running the `npm run format` rule.

PR #961, opened 2023-06-27
A11y Bug Fix: Auth User Table Dropdown Context

ABradham:
Addresses b/262062452.

Adds context to dropdown menu for Editing, Deleting, or Disabling users on the Authentication `Users` table. With this change, either the `displayName`, `email`, or `phoneNumber` (one of these must exist for a user to be created in the table) of the user is read by the screenreader so that users of this assistive technology will know which user is being edited/disabled/deleted.

### DIFF
--- a/src/components/Auth/UsersCard/table/UsersTable.tsx
+++ b/src/components/Auth/UsersCard/table/UsersTable.tsx
@@ -91,7 +91,8 @@ export const UsersTableRow: React.FC<
             <IconButton
               theme="secondary"
               icon="more_vert"
-              label={`Open menu for user ${
+              label="Open menu"
+              aria-label={`Open menu for user ${
                 user.displayName || user.email || user.phoneNumber
               }`}
             />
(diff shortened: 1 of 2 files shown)


src/components/Auth/UsersCard/table/UsersTable.tsx
```diff
@@ -88,7 +88,14 @@ export const UsersTableRow: React.FC<
           onOpen={() => setMenuOpen(true)}
           onClose={() => setMenuOpen(false)}
           handle={
-            <IconButton theme="secondary" icon="more_vert" label="Open menu" />
+            <IconButton
+              theme="secondary"
+              icon="more_vert"
+              aria-label={`Open menu for user ${
+                user.displayName || user.email || user.phoneNumber
```
christhompsongoogle, line 95: Let's leave the phone number off - unless we prepend it with something that indicates it's a phone number.

christhompsongoogle (changes requested):
Notes: Adjust label to remove phone number

src/components/Auth/UsersCard/table/UsersTable.tsx
ABradham: Gotcha; I'll add some additional text for when a user's Identifier is  a phone number

christhompsongoogle (approved):

PR #963, opened 2023-06-28
Fixes an issue with double tab targets in the Extensions page.

christhompsongoogle:
Previously the link target had a nested button. Making either one `aria-hidden` resulted in incorrect behavior. Repurposed the button as a link-button.

Internal: b/259456661 and b/259457073


ABradham (approved):
LGTM!

Edit: The test failure here seems to be that weird race condition thing we discovered on Wednesday and doesn't seem to be related to the change you made

PR #962, opened 2023-06-28
A11y Bug Fix: Firestore delete document field dialog

ABradham:
Addresses bug b/262062132.

Adds a confirmation dialog when the user tries to delete a document field in the firestore database ui.

PR #965, opened 2023-06-30
A11y Bugfix: Add confirmation dialog on deleting `1..n` selected items from storage

ABradham:
Adds a confirmation dialog when a user attempts to delete `n` files from storage. Matches behavior of https://github.com/firebase/firebase-tools-ui/pull/962.
 
Internal: Addresses b/263429461

### DIFF
--- a/src/components/Storage/Card/Table/Header/ActionHeader/ActionHeader.tsx
+++ b/src/components/Storage/Card/Table/Header/ActionHeader/ActionHeader.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021 Google LLC
+ * Copyright 2023 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/components/Storage/Card/Table/Header/ActionHeader/ActionHeader.tsx
```diff
@@ -75,9 +76,11 @@ export const ActionHeader: React.FC<
               Open all files
             </Button>
             <Button
-              onClick={() => {
-                selection.clearAll();
-                return deleteFiles(paths);
+              onClick={async () => {
+                if (await confirmDeleteSelectedFiles()) {
+                  selection.clearAll();
+                  return deleteFiles(paths);
```
christhompsongoogle, line 82: We'll need to update the test here:
https://github.com/firebase/firebase-tools-ui/blob/bfff08543abf3fd62b36eda2a1f694567699859f/src/components/Storage/Card/Table/Header/ActionHeader/ActionHeader.test.tsx#L23

christhompsongoogle (approved):

PR #979, opened 2023-07-21
Experiment flags

jhuleatt:
Allow experiments set with `firebase experiments:enable` to be accessed by the emulator UI. Corresponding CLI PR: https://github.com/firebase/firebase-tools/pull/6169

The new `useExperiment()` hook will return a list of enabled experiments, so any component that wants to conditionally show UI based on an experiment flag can call this hook:

```jsx
function ExperimentalFeatureUI() {
  const showNewFeature = useExperiment("pineapple-smoothie");

  if (showNewFeature === true) {
    return <h1>Hi, I am an experimental feature</h1>;
  } else {
    return null;
  }
}
```

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -53,27 +53,20 @@ app.get(
   '/api/config',
   jsonHandler(async () => {
     const hubDiscoveryUrl = new URL(`http://${hubHost}/emulators`);
-    const experimentDiscoveryUrl = new URL(
-      `http://${hubHost}/enabled-experiments`
-    );
+    const emulatorsRes = await fetch(hubDiscoveryUrl.toString());
+    const emulators = (await emulatorsRes.json()) as any;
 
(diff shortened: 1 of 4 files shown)


server.ts
```diff
@@ -53,10 +53,22 @@ app.get(
   '/api/config',
   jsonHandler(async () => {
     const hubDiscoveryUrl = new URL(`http://${hubHost}/emulators`);
-    const emulatorsRes = await fetch(hubDiscoveryUrl.toString());
-    const emulators = (await emulatorsRes.json()) as any;
+    const experimentDiscoveryUrl = new URL(
+      `http://${hubHost}/enabled-experiments`
```
kroikie, line 57: Does this endpoint still exist on the CLI side?

jhuleatt: Updated to remove this call and use env var instead

yuchenshi (approved):

christhompsongoogle (approved):

PR #998, opened 2023-11-27
Multi db url

christhompsongoogle:


package.json
```diff
@@ -228,4 +234,4 @@
   "eslintConfig": {
     "extends": "react-app"
   }
-}
+}
```
joehan, line 237: newline

src/components/Firestore/CollectionFilter/CollectionFilter.test.tsx
```diff
@@ -94,6 +94,13 @@ it('unsets sorting when switching to an operator that does not support it', asyn
 });
 
 it('supports multi-value operators', async () => {
+    // Stub the browser property performance.now()
+    Object.defineProperty(window, 'performance', {
```
joehan, line 98: Not sure exactly how these tests run, so this might be misguided, but double check that this does not leak into other tests

src/components/Firestore/index.tsx
```diff
@@ -152,8 +155,7 @@ export const Firestore: React.FC<React.PropsWithChildren<unknown>> = React.memo(
               </div>
             }
           >
-            The Emulator Suite UI only supports the (default) database right
-            now.
+            The Emulator Suite UI supports multiple DBs by editing the database name in the URL.
```
joehan, line 158: Totally reasonable for now, but this does seem like low hanging fruit to provide a UI component to choose from a list of existing DBs or 'create' a new one by typing in a databaseId and redirecting to the approriate url.

joehan (approved):
LGTM with a few small things

src/components/Firestore/index.tsx
christhompsongoogle: 100% we should do this - all it takes it a dropdown or something

src/components/Storage/index.test.tsx
```diff
@@ -29,7 +29,7 @@ describe('Storage Index Page', () => {
     const { getByText, defaultBucket } = await renderWithStorage(
       <StorageWrapper />
     );
-    await waitFor(() => delay(100));
+    await waitFor(() => delay(200));
```
christhompsongoogle, line 32: Got a few flakes, increased delay a bit to help out.

yuchenshi (approved):

src/components/Firestore/CollectionFilter/CollectionFilter.test.tsx
christhompsongoogle: Ack - did some testing and it doesn't seem to affect other tests - I think this property is mostly unused as this is the only test that requires it

PR #1052, opened 2024-10-01
Adding SDK-created extensions support

ifielker:
Adding SDK-created extensions support

### DIFF
--- a/src/components/Extensions/Details/DetailCard/Tabs/ParamList.tsx
+++ b/src/components/Extensions/Details/DetailCard/Tabs/ParamList.tsx
@@ -21,52 +21,51 @@ import { Callout } from '../../../../common/Callout';
 import { DOCS_BASE } from '../../../../common/links/DocsLink';
 import { Markdown } from '../../../../common/Markdown';
 import { useExtension } from '../../../api/useExtension';
+import { isDynamicExtension } from '../../../api/useExtensions';
 import { EventsConfig } from './EventsConfig';
 import styles from './ParamList.module.scss';
 import { ParamValue } from './ParamValue';
-import { isDynamicExtension } from '../../../api/useExtensions';
 
 function ParamList() {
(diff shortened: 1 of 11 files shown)


ifielker:
Added extension instance id in the list
image

Changed message for non-SDK-created extensions
image

New message for SDK-created extensions
image

Message no longer specific to non-SDK-created extensions (now includes both)
image

src/components/Extensions/Details/DetailCard/Tabs/ParamList.tsx
```diff
@@ -21,36 +21,51 @@ import { Callout } from '../../../../common/Callout';
...
-              >
-                <Button>Learn more</Button>
-              </a>
-            </div>
-          </Accordion>
+          {isDynamic ? dynamicFragment : staticFragment}
+          <div className={styles.learnButton}>
+            <a
+              href={DOCS_BASE + docSuffix}
+              target="_blank"
```
tjlav5, line 62: You can use the ExternalLink component here, also why a tab-index of -1?

src/components/Extensions/Details/DetailCard/Tabs/ParamList.tsx
```diff
@@ -21,36 +21,51 @@ import { Callout } from '../../../../common/Callout';
...
+              rel="noopener noreferrer"
+              tabIndex={-1}
+            >
+              <Button>Learn more</Button>
```
tjlav5, line 66: Hmm I don't think we should have buttons inside of anchors... that's not valid HTML

tjlav5 (approved):

src/components/Extensions/Details/DetailCard/Tabs/ParamList.tsx
ifielker: discussed offline. This is ok.

PR #1077, opened 2025-04-22
Starting to make this work in Studio

joehan:


avolkovi (approved):

src/components/common/useRequest.ts
```diff
@@ -53,3 +55,7 @@ export function useRequest<T = unknown>(
     refreshWhenOffline: true,
   });
 }
+
+function isCloudWorkstation(url: string) {
+  return url.includes('cloudworkstations.dev');
```
maneesht, line 60: I think this should be `.endsWith`
```suggestion
  return url.endsWith('.cloudworkstations.dev');
```

src/firebase.ts
```diff
@@ -64,3 +75,13 @@ export function useEmulatedFirebaseApp(
 
   return app;
 }
+
+const loadCloudWorkstationCookies = async (url?: string) => {
+  if (url && url.includes('cloudworkstations.dev')) {
```
maneesht, line 80: Can we just re-use the `isCloudWorkstation` call here?

src/firebase.ts
```diff
@@ -64,3 +75,13 @@ export function useEmulatedFirebaseApp(
...
+    try {
+      const res = await fetchMaybeWithCredentials(url);
```
maneesht, line 82: Something I'm considering down the line is exponential backoff for the situation when the user hasn't started the emulator yet

package.json
```diff
@@ -42,7 +42,7 @@
     "classnames": "^2.2.6",
     "codemirror": "^5.61.1",
     "express": "^4.17.1",
-    "firebase": "^10.7.1",
+    "firebase": "^11.6.1-firebase-studio-sdk-integration.226be0bb1",
```
maneesht, line 45: Are we merging this for an official release?

src/components/Firestore/FirestoreEmulatedApiProvider.tsx
```diff
@@ -54,6 +54,9 @@ export const FirestoreEmulatedApiProvider: React.FC<
     useCallback(
       (app: FirebaseApp) => {
         const firestore = getFirestore(app, databaseId);
+        console.log(
```
maneesht, line 57: Do we need to keep this console.log?

src/components/common/EmulatorConfigProvider.tsx
```diff
@@ -234,6 +234,10 @@ async function configFetcher(url: string): Promise<Config> {
           // resolve it to IPv6 and connection may fail.
           host = '127.0.0.1';
         }
+        if (key == 'firestore') {
+          // If we remapped the host, apply it to the websocket host too.
+          value.webSocketHost = value.host;
```
maneesht, line 239: should this be host instead of `value.host`?

src/components/common/EmulatorConfigProvider.tsx
```diff
@@ -242,21 +246,21 @@ async function configFetcher(url: string): Promise<Config> {
           // Ditto for IPv6.
           host = '::1';
         }
+        if (key == 'firestore') {
+          // If we remapped the host, apply it to the websocket host too.
+          value.webSocketHost = value.host;
```
maneesht, line 251: Same question here

src/components/common/EmulatorConfigProvider.tsx
```diff
@@ -242,21 +246,21 @@ async function configFetcher(url: string): Promise<Config> {
...
+
     result[key as Emulator] = {
       ...value,
       host,
       port,
       hostAndPort: hostAndPort(host, port),
     };
   }
-
-  if (result.firestore?.webSocketPort) {
```
avolkovi, line 256: previously you only remapped the webSocketPort if it was set, do you need to preserve this if statement?

src/components/common/rest_api.ts
```diff
@@ -115,3 +115,23 @@ function toFormData(file: File) {
   formData.append('upload', file);
   return formData;
 }
+
+function isCloudWorkstation(url: string) {
+  return url.includes('cloudworkstations.dev');
```
avolkovi, line 120: let's prepend with a "\." so we don't accidentally catch other domains that just happen to end in this string, see @maneesht comment in the other implementation of this

src/components/common/useRequest.ts
```diff
@@ -53,3 +55,7 @@ export function useRequest<T = unknown>(
     refreshWhenOffline: true,
   });
 }
+
+function isCloudWorkstation(url: string) {
```
avolkovi, line 59: this is also defined in rest_api, did you mean to have two implementations?

src/firebase.ts
avolkovi: ➕

avolkovi (approved):

src/components/common/useRequest.ts
joehan: Nope, url can include paths - ie: '9099-blah.cluster.cloudworkstations.dev/auth'

package.json
joehan: No, we'll wait for the real release - adding a note to the descriptio n

src/components/common/EmulatorConfigProvider.tsx
joehan: Good catch, done

src/components/common/EmulatorConfigProvider.tsx
joehan: I think it's likely irrelevant (webSocketPort is always provided), but added it back in

src/components/common/useRequest.ts
joehan: Nope, this one was actually unused - i should be more careful with self review lol

src/components/common/rest_api.ts
```diff
@@ -115,3 +115,23 @@ function toFormData(file: File) {
...
+// Previously, we had a ton a raw fetch calls throughout the codebase, so any broad changes needed to be made all over the place.
+export async function fetchMaybeWithCredentials(
+  url: string,
+  opts?: RequestInit
+): Promise<Response> {
+  if (!opts) {
+    opts = {};
+  }
+  if (isCloudWorkstation(url)) {
+    opts.credentials = 'include';
```
yuchenshi, line 134: I think you can just do

```suggestion
    opts = {...opts, credentials: 'include'};
```
It will handle undefined without Line 130-132 and it avoids modifying the argument

src/firebase.ts
```diff
@@ -40,9 +44,20 @@ export function useEmulatedFirebaseApp(
   config?: object,
   initialize?: (app: FirebaseApp) => void
 ): FirebaseApp | undefined {
-  const { projectId } = useConfig();
+  const conf = useConfig();
+  const projectId = conf.projectId;
   const [app, setApp] = useState<FirebaseApp | undefined>();
-
+  void Promise.all([
```
yuchenshi, line 50: nit: I think you can just void each without Promise.all

yuchenshi (approved):

src/components/common/EmulatorConfigProvider.tsx
maneesht: I don't see the update here. Did you forget to push?

src/components/common/useRequest.ts
maneesht: nit: I think my preference is that we call this with the hostname path if we have it and therefore have `endsWith` just so that it's a stronger validation

package.json
```diff
@@ -76,7 +76,7 @@
   },
   "overrides": {
     "reactfire": {
-      "firebase": "^10.7.1"
+      "firebase": "^11.6.0"
```
maneesht, line 79:
```suggestion
      "firebase": "^11.7.0"
```

src/components/common/EmulatorConfigProvider.tsx
maneesht: Did you mean to bring this back?

yuchenshi (approved):

